PR #8, opened 2020-07-01
Set ovn_central_ip to point to all central IPs by default.

dceara:
By default we start all tests with NB/SB DBs clustered. Make
configuration less error prone by setting ovn_central_ip to point to all
central IPs by default.


numansiddique:
Acked-by: Numan Siddique <numans@ovn.org>

PR #9, opened 2020-07-06
improve code to support rhel

linup2011:


dceara:
This depends on https://github.com/ovn-org/ovn-fake-multinode/pull/29 so I'll wait with review until the ovn-fake-multinode PR gets merged.

ovn-fake-multinode-utils/playbooks/install-dependencies.yml
```diff
@@ -2,13 +2,7 @@
...
-        name:
-          - git
-          - docker
-          - docker-distribution
-          - gcc
-          - openvswitch
-          - python3-pyyaml
-          - python3-virtualenv
-          - python3-devel
+      shell: yum install git  gcc openvswitch python3-pyyaml python3-virtualenv python3-devel  --skip-broken -y
```
dceara, line 5: Nit: extra whitespace between `git` and `gcc`.

ovn-fake-multinode-utils/playbooks/install-dependencies.yml
```diff
@@ -2,13 +2,7 @@
...
+    - name: Install container command
+      shell: yum install docker docker-distribution -y ||  yum install podman podman-docker -y
```
dceara, line 7: Nit: extra whitespace before `yum install podman podman-docker -y`.

do.sh
```diff
@@ -163,13 +198,52 @@ function install_ovn_fake_multinode() {
 
     # Clone repo locally
     clone_component ovn-fake-multinode ${ovn_fmn_repo} ${ovn_fmn_branch} || rebuild_needed=1
-    clone_component ovs ${ovs_repo} ${ovs_branch} || rebuild_needed=1
-    clone_component ovn ${ovn_repo} ${ovn_branch} || rebuild_needed=1
+
+    if [ -n "$RPM_OVS" ]
```
dceara, line 202: Do you mind adding a section in the readme about these new variables that allow deployments with prebuilt RPMs? I.e., documenting `RPM_OVS`, `RPM_OVN_COMMON`, `RPM_OVN_CENTRAL`, `RPM_OVN_HOST`, `RPM_SELINUX`.

do.sh
```diff
@@ -45,14 +48,46 @@ function install_deps() {
...
+    echo "-- Configuring podman local registry on tester nodes"
+    if which podman
+    then
+        for container_name in `podman ps | grep -v "CONTAINER ID" | awk '{print $1}'`
+        do
+            podman stop $container_name
+            podman rm $container_name
+        done
+        [ -d /var/lib/registry ] || mkdir /var/lib/registry -p
+        podman run --privileged -d --name registry -p 5000:5000 \
```
dceara, line 72: Shouldn't this part be moved to `install_deps()` where we install and start local dependencies? Also, I don't think we need `registries = ['localhost:5000']` on the tester node, right?

linup2011: I added this in case someone use the same system as both tester node and worker node

do.sh
linup2011: sure, I will

do.sh
linup2011: and this is used to configure podman, so added it here. should I remove it to install_deps?

dceara: Yes, I think everything that's related to the "tester" node should go to `install_deps()` in the "-- Installing local dependencies" section.

linup2011:
updated

PR #11, opened 2020-07-08
do.sh: Clone projects into correct directories.

blp:
"git clone" uses the repo name as the default name for the directory
it creates.  This is right for the default repos, but users might
override the repo with one that has a different name, in which case
the default will be wrong.

By specifying the directory name explicitly like this, it will always
work properly.

I had this problem when I tried to use my ovs-reviews repo for OVS
and OVN.

Signed-off-by: Ben Pfaff <blp@ovn.org>


numansiddique:
Acked-by: Numan Siddique <numans@ovn.org>

dceara:
Thanks @blp for the fix!

PR #17, opened 2020-08-25
do.sh: Use docker-ce when available.

dceara:
This enables users to use docker on platforms where docker is not
officially supported, e.g., RHEL8.

Also, remove dependency on docker-distribution. Instead just start a
container that runs the registry on the tester node.


numansiddique (approved):

numansiddique:
LGTM.

PR #22, opened 2020-12-11
Collect information about running processes at test completion.

igsilya:
It's good to know what processes are running with their exact command
lines and how many resources they're using.  Fr example, it's very
important to know what is the current memory consumption per process in
the end of the test, e.g. if you're working on memory related
optimizations.


dceara (approved):
LGTM, thanks!

PR #30, opened 2021-06-22
ovn-tester: Refactor initial code.

dceara:
Note: There are still a bunch of "FIXME" in the code that need to be
addressed to make writing tests easier.  Also the configuration parsing
in ovn_tester.py along with the way tests are run, could be improved.

Added TODO file to track the above (and some more).

Otherwise this commit:
- refactors SSH connections, removes RemoteConn, instead we now define
  objects that correspond to each physical node and only create on SSH
  connection for each of those.
- defines CentralNode and WorkerNode, which are containerized OVN nodes
  running on top of PhysicalNode.  Multiple containerized OVN nodes may
  run on the same PhysicalNode.
- refactors ovn_workload.py to make it a bit more object oriented and to
  avoid passing dictionaries with random configuration knobs here and
  there.
- renames class OvnWorkload to Cluster because that's what it actually
  represents.
- removes quite a bit of all the configuration knobs inherited from
  ovn-scale-test because we don't need most of them (at least for now).
- changes all sample configs from test-scenarios/*.yml to match the
  new code.


putnopvut (approved):
For a partial refactor, this works just fine. I think there are some nitpicky comments I could make ("use f-strings!!!") but it's not worth doing that until a larger refactor is performed.

dceara:
> For a partial refactor, this works just fine. I think there are some nitpicky comments I could make ("use f-strings!!!") but it's not worth doing that until a larger refactor is performed.

Thanks!  Yes, more refactoring is still needed, this was just the ground work.  I'll merge this.

PR #31, opened 2021-06-23
Add Load Balancer API and use it to create cluster load balancer during cluster bring-up

putnopvut:
This is part one of a series that seeks to update the bring-up scenario to be closer to our scale testing goals.

In this series of commits:
Commit 1: Introduces a Load Balancer API.
Commit 2: Uses the API to create a cluster load balancer and add some initial VIPs to it.

These commits do not add per-node load balancers, nor do they update the cluster load balancers after initial creation. This is because the refactor that this patch is based off does not have the node-specific bring-up defined. This is best left until the rest of the code has been refactored.


dceara:
Thanks!

PR #32, opened 2021-06-23
ovn-tester: Add multitenant network policy-like tests.

dceara:
Allows scale testing OpenShift multitenant network policy-like
scenarios, for example, the equivalent of:

```
for i in range(n_namespaces):
    create address set AS_ns_i
    create port group PG_ns_i
    if i < 200:
        n_pods = 1 # 200 pods
    elif i < 480:
        n_pods = 5 # 1400 pods
    elif i < 495:
        n_pods = 20 # 300 pods
    else:
        n_pods = 100 # 500 pods
    create n_pods
    add n_pods to AS_ns_i
    add n_pods to PG_ns_i
    create acls:

    to-lport, ip.src == $AS_ns_i && outport == @PG_ns_i, allow-related
    to-lport, ip.src == {ip1, ip2, ip3} && outport == @PG_ns_i, allow-related
    to-lport, ip.src == {ip1, ..., ip20} && outport == @PG_ns_i, allow-related
```

### DIFF
--- a/ovn-tester/ovn_tester.py
+++ b/ovn-tester/ovn_tester.py
@@ -12,7 +12,7 @@ from collections import namedtuple
 from ovn_context import Context
 from ovn_utils import PhysicalNode
 from ovn_workload import BrExConfig, ClusterConfig
-from ovn_workload import CentralNode, WorkerNode, Cluster
+from ovn_workload import CentralNode, WorkerNode, Cluster, Namespace
 
 DEFAULT_VIP_SUBNET = netaddr.IPNetwork('4.0.0.0/8')
 DEFAULT_N_VIPS = 2
@@ -46,6 +46,15 @@ def calculate_default_static_vips():
 
 ClusterBringupCfg = namedtuple('ClusterBringupCfg',
                                ['n_pods_per_node'])
+
+
+NsRange = namedtuple('NsRange',
+                     ['start', 'n_pods'])
+NsMultitenantCfg = namedtuple('NsMultitenantCfg',
+                              ['n_namespaces',
+                               'ranges',
+                               'n_external_ips1',
+                               'n_external_ips2'])
 
 
 def usage(name):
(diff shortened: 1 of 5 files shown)


ovn-tester/ovn_tester.py
```diff
@@ -118,7 +127,21 @@ def read_config(configuration):
...
+        netpol_multitenant_args = config.get('netpol_multitenant', dict())
+        netpol_multitenant_cfg = NsMultitenantCfg(
+            n_namespaces=netpol_multitenant_args.get('n_namespaces', 0),
+            n_external_ips1=netpol_multitenant_args.get('n_external_ips1', 3),
+            n_external_ips2=netpol_multitenant_args.get('n_external_ips2', 20),
+            ranges=[
+                NsRange(
+                    start=range_args.get('start', 0),
+                    n_pods=range_args.get('n_pods', 5),
+                ) for range_args in netpol_multitenant_args.get('ranges')
```
putnopvut, line 140: What happens here if `netpol_multenant_args` has no `ranges` configured? I expect that `netpol_multitenant_args.get()` will return `None`, resulting in a `TypeError`.

ovn-tester/ovn_tester.py
```diff
@@ -153,27 +176,74 @@ def run_base_cluster_bringup(ovn, bringup_cfg):
...
+        str(netaddr.IPAddress('43.43.43.1') + i)
+        for i in range(cfg.n_external_ips2)
+    ]
+
+    with Context("netpol_multitenant", cfg.n_namespaces) as ctx:
+        for i in ctx:
+            n_ports = 1
+            for r in cfg.ranges:
+                if i >= r.start:
+                    n_ports = r.n_pods
```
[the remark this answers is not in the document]
putnopvut: Alternately, you could sort `cfg.ranges` in descending order and break as soon as `i  >= r.start`, potentially saving some loop iterations.

ovn-tester/ovn_utils.py
```diff
@@ -159,21 +160,21 @@ def ls_port_add(self, lswitch, name="", router_port=None,
...
+        elif mac or ip:
             cmd = "lsp-set-addresses {} \"".format(name)
-            if len(mac):
-                cmd = cmd + mac + " "
-            if len(ip):
-                cmd = cmd + ip
+            if mac:
+                cmd = cmd + str(mac) + " "
+            if ip:
+                cmd = cmd + str(ip)
```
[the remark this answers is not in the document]
dceara: Done.

ovn-tester/ovn_tester.py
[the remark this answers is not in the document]
dceara: Done, thanks for the suggestion.

ovn-tester/ovn_tester.py
dceara: Fixed, thanks for spotting this!

ovn-tester/ovn_tester.py
```diff
@@ -118,7 +128,23 @@ def read_config(configuration):
...
-        return log_cmds, cluster_cfg, brex_cfg, bringup_cfg
+
+        netpol_multitenant_args = config.get('netpol_multitenant', dict())
+        ranges=[
+            NsRange(
+                start=range_args.get('start', 0),
+                n_pods=range_args.get('n_pods', 5),
+            ) for range_args in netpol_multitenant_args.get('ranges', list())
+        ]
+        ranges.sort(key=operator.itemgetter(0), reverse=True)
```
putnopvut, line 139: This might be a bit more readable as:
```
ranges.sort(key=operator.attrgetter('start'), reverse=True)
```
This at least makes it more clear that the `start` attribute is the key.
Or if you want to avoid importing `operator`, then you could go with:
```
ranges.sort(key=lambda x: x.start, reverse=True)
```

dceara: I went for `ranges.sort(key=lambda x: x.start, reverse=True)`, thanks for the suggestion.

PR #33, opened 2021-06-23
Ovn tester refactor

dceara:
- Use f-strings.
- Don't use raw dictionaries, use namedtuple instead.
- Some more minor fixes.
- Factor out some modules from ovn_utils.py

Also, make NB/SB raft election timer configurable.


dceara:
Based on internal discussion with @LorenzoBianconi, merging this.

PR #38, opened 2021-06-25
Cleanup resources and add OpenShift specific test configs.

dceara:
Also:
- fix all pep8 reported warnings.
- don't enforce the default drop policy in namespaces unless an explicit
  policy is added.
- by default skip the cleanup stage (configurable through
  global.cleanup).
- make ovn-nbctl always connect to RAFT leader.


LorenzoBianconi:

PR #44, opened 2021-07-01
ovn-tester: Create namespace port groups only when enforcing policies.

dceara:
This simulates what ovn-kubernetes does: port groups are created only
when the first network policy is added to the namespace.  Address sets
are created from the beginning however.


LorenzoBianconi:

PR #45, opened 2021-07-01
ovn-tester: Set NB/SB inactivity_probe.

dceara:
Make the DB inactivity probe configurable and set it to 60 seconds by
default to mimic what OpenShift does.


LorenzoBianconi:

PR #48, opened 2021-07-08
Stop using ovn-scale-test (rally-ovs) and browbeat.

dceara:


dceara:
Discussed with @LorenzoBianconi offline, will merge this.

PR #50, opened 2021-07-09
ovn-tester: Modularize test definitions.

putnopvut:
This change encapsulates tests into their own files. Each yaml section
in the configuration file corresponds to a module in the tests/
directory. This has the following advantages:

* The logic of individual tests can be contained within its own
  module/class.
* The ovn_tester.py file can be made smaller, allowing for it to be the
  orchestrator of tests, rather than holding test logic.
* New tests can be defined without needing to modify ovn_tester.py. Just
  add the new module to tests/ and as long as the configuration is
  present, the new test will be run.
* This allows for customized ordering of test scenarios.
* This allows for test scenarios to be skipped if desired.

In this change, the cluster bringup test is handled in a special way. It
is run directly in ovn_tester.py since it must be run as the first test.
This could be made a bit less awkward by changing the config file
schema, but I sought to make this change not break things for people if
possible.

### DIFF
--- a/ovn-tester/ovn_tester.py
+++ b/ovn-tester/ovn_tester.py
@@ -7,6 +7,7 @@ import ovn_stats
 import netaddr
 import time
 import yaml
+import importlib
 
 from collections import namedtuple
 from ovn_context import Context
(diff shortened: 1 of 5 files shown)


ovn-tester/ovn_tester.py
```diff
@@ -88,97 +72,78 @@ def read_physical_deployment(deployment, global_cfg):
...
+            'cluster_cmd_path',
+            '/root/ovn-heater/runtime/ovn-fake-multinode'
+        ),
+        monitor_all=cluster_args.get('monitor_all', True),
+        logical_dp_groups=cluster_args.get('logical_dp_groups', True),
+        clustered_db=cluster_args.get('clustered_db', True),
+        raft_election_to=cluster_args.get('raft_election_to', 16),
+        node_net=netaddr.IPNetwork(
+            cluster_args.get('node_net', '192.16.0.0/16')
+        ),
```
dceara, line 94: This got removed by accident I guess.
```
db_inactivity_probe=cluster_args.get('db_inactivity_probe', 60000),
```

putnopvut: Whoops. Good spot. I'll push a fix.

dceara:
Looks good to me, thanks!

PR #54, opened 2021-07-15
log-collector: Collect process stats before and after compaction.

igsilya:
Process stats before and after compaction will allow to see how much
memory the process actually uses removing the noise caused by glibc
not releasing fastbins back to the system. (This is true, because
memory trimming is enabled on compaction).  Stats 'before' are useful
to estimate possible runtime memory consumption.

Also, compaction will save disc space by reducing size of the database
files.


dceara:
Looks good to me, thanks!

PR #60, opened 2021-07-20
ovn-tester: Set northd probe interval.

dceara:
Default 5000ms (as currently configured by OpenShift).


LorenzoBianconi:

PR #62, opened 2021-07-23
Integrate ovsdb-etcd to ovn-heater and create etcd test scenario.

msherif1234:
Signed-off-by: Mohamed S. Mahmoud <mmahmoud@redhat.com>

### DIFF
--- a/ovn-tester/ovn_tester.py
+++ b/ovn-tester/ovn_tester.py
@@ -47,7 +47,7 @@ ClusterBringupCfg = namedtuple('ClusterBringupCfg',
                                ['n_pods_per_node'])
 
 
-def calculate_default_node_remotes(net, clustered, n_relays):
+def calculate_default_node_remotes(net, clustered, n_relays, enable_ssl):
     ip_gen = net.iter_hosts()
     if n_relays > 0:
         skip = 3 if clustered else 1
@@ -56,7 +56,10 @@ def calculate_default_node_remotes(net, clustered, n_relays):
         ip_range = range(0, n_relays)
     else:
         ip_range = range(0, 3 if clustered else 1)
-    remotes = ["ssl:" + str(next(ip_gen)) + ":6642" for _ in ip_range]
+    if enable_ssl:
+        remotes = ["ssl:" + str(next(ip_gen)) + ":6642" for _ in ip_range]
+    else:
+        remotes = ["tcp:" + str(next(ip_gen)) + ":6642" for _ in ip_range]
     return ','.join(remotes)
 
 
@@ -93,6 +96,7 @@ def read_config(config):
     clustered_db = cluster_args.get('clustered_db', True)
     node_net = netaddr.IPNetwork(
         cluster_args.get('node_net', '192.16.0.0/16'))
+    enable_ssl = cluster_args.get('enable_ssl', True)
     n_relays = cluster_args.get('n_relays', 0)
     cluster_cfg = ClusterConfig(
         cluster_cmd_path=cluster_args.get(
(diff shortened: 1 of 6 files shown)


ovn-tester/ovn_tester.py
```diff
@@ -114,7 +114,9 @@ def read_config(config):
         vips=cluster_args.get('vips', calculate_default_vips()),
         vip_subnet=DEFAULT_VIP_SUBNET,
         static_vips=cluster_args.get('static_vips',
-                                     calculate_default_static_vips())
+                                     calculate_default_static_vips()),
+        use_ovsdb_etcd=cluster_args.get('use_ovsdb_etcd', False),
+        enable_ssl=cluster_args.get('enable_ssl', False)
```
dceara, line 119: SSL used to be enabled by default, let's leave it like that.

physical-deployments/physical-deployment.yml
```diff
@@ -1,9 +1,8 @@
-registry-node: host01.mydomain.com
-internal-iface: eno1
+registry-node: localhost
```
dceara, line 1: I guess the physical-deployment.yml change was committed by accident.

test-scenarios/ovn-etcd-low-scale.yml
```diff
@@ -0,0 +1,35 @@
+global:
+  log_cmds: True
```
dceara, line 2: Probably better to make it less verbose by disabling command logging.

test-scenarios/ovn-low-scale.yml
```diff
@@ -3,6 +3,7 @@ global:
 
 cluster:
   clustered_db: true
+  enable_ssl: False
```
dceara, line 6: This seems unrelated, please revert this.

dceara (changes requested):
Thanks for the contribution, I left a few comments.  Also, it would be great to mention what dependencies (e.g., ovn-fake-multinode, ovsdb-etcd) this has.

test-scenarios/ovn-low-scale.yml
msherif1234: > This seems unrelated, please revert this.

its very related for ovs-etcd we need to use tcp protocol since the client only listen to tcp on port 6641

test-scenarios/ovn-etcd-low-scale.yml
```diff
@@ -0,0 +1,36 @@
+global:
+  log_cmds: False
+
+cluster:
+  cluster_cmd_path: /home/mmahmoud/ovn-heater/runtime/ovn-fake-multinode
```
msherif1234, line 5: @dceara I need the above setting w/o containers won't start and in the same time I can't leave like that what are you suggestions ?

dceara: @msherif1234 Sure, but that can be a local change, in your repo.  These files are "samples" of test scenarios and should probably be as generic as possible.  Ideally we should get rid of `cluster_cmd_path` completely but that's a longer story and is not relevant for this PR.

dceara:
@msherif1234 There are also a few failed [flake8 checks](https://github.com/dceara/ovn-heater/runs/3173914118?check_suite_focus=true), could you please fix those too?

ovn-tester/ovn_workload.py
```diff
@@ -117,10 +122,15 @@ def start(self, cluster_cfg):
     @ovn_stats.timeit
     def connect(self, cluster_cfg):
         print(f'***** connecting worker {self.container} *****')
+        if not cluster_cfg.enable_ssl:
+            node_remote = \
+                'tcp:192.16.0.1:6642,tcp:192.16.0.2:6642,tcp:192.16.0.3:6642'
```
dceara, line 127: I think the best way to do this is to actually change calculate_default_node_remotes in ovn_tester.py and pass it enable_ssl based on which it should either set tcp or ssl as protocol [here](https://github.com/dceara/ovn-heater/blob/6aeafc0ec64296cf7f018700fed9df8904fa482d/ovn-tester/ovn_tester.py#L59).

dceara (changes requested):

dceara (approved):
Looks good to me, thanks! I'm waiting for the [ovn-fake-multinode PR](https://github.com/ovn-org/ovn-fake-multinode/pull/46) to merge first before pushing this one.

PR #66, opened 2021-08-11
Convert ovn-tester to use asyncio

putnopvut:
ovn-tester currently works in a synchronous and serialized manner. If we are going to provision, say, 50 ports, the operation will go something like:
- Issue SSH command to create a port, and wait for that to return
- Issue SSH commands to update load balancers, and wait for them to return
- Issue SSH commands to ping ports, and wait for them to return
Repeat 50 times.

With this series, we use python's asyncio framework to create tasks that are each responsible for performing the above operations. Each task looks something like:
- Issue SSH command to create a port, then yield control to the event loop
- Collect result of SSH command, then issue SSH commands to update load balancers. Yield control to the event loop.
- Collect result of SSH command, then issue SSH commands to ping ports. Yield control to the event loop.
- Collect result of SSH command.
With asyncio, we can deploy multiple of these tasks at once, allowing the event loop to give control to the individual tasks as they have data to work with. The tests have been updated to take a "queries_per_second" option that allows for the number of tasks to create per second to be configured. By default, all the tests create 20 tasks per second.

In order to support asyncio, this also changes the SSH library from paramiko to asyncssh.

See the individual commits in this PR for explanations of all of what is happening in the code.

Some notes:
- This change makes use of asyncio functions that are only available in python 3.7+. ovn-heater should probably be updated to ensure that a proper version of python is used in the virtualenv when running tests.
- When testing this change in a cluster of RHEL8 servers, I found that the central container that was receiving the ovn-nbctl commands would get hung. It appeared that issuing multiple `docker exec` calls at once caused the container to deadlock. I fixed this by upgrading the docker version to 20.10.8 on the server in question. ovn-heater should probably be updated to ensure that a new enough docker version is installed on the physical nodes.
- Stats are accurate, but may potentially be misleading.
  - Since test iterations run concurrently, adding together multiple iterations' times doesn't make much sense. As an example, I ran the `ovn-20.yml` scenario on a cluster, and the entire test took 20 minutes to complete. However, looking at the `netpol_multitenant` report, the sum of all iterations is 70 hours. 
  - Measurements include idle time while the event loop runs other tasks. As an example, let's say we issue an SSH command to ping a port. 5 seconds later, the SSH command completes. However, because the event loop is running other tasks, the result of that SSH command is not collected until 30 seconds after the ping was initially issued. In this case, the stats would show the ping took 30 seconds. It may be worth changing our measurements to be performed on the server instead (or in addition to our client-side timing)


ovn-tester/ovn_workload.py
```diff
@@ -528,83 +548,90 @@ def __init__(self, central_node, worker_nodes, cluster_cfg, brex_cfg):
...
         )
 
-    def provision_ports(self, n_ports, passive=False):
-        return [
-            self.select_worker_for_port().provision_ports(self, 1, passive)[0]
-            for _ in range(n_ports)
-        ]
+    async def provision_ports(self, n_ports, passive=False):
+        ret_list = []
+        for _ in range(n_ports):
```
mohammadheib, line 610: HI Mark these changes looks very good to me but here maybe we can create ports concurrently using asyncio.gather rather than await i think gather will consume less time if we create list of coroutines to create list of ports at the same time : (please see the suggested change below):

-        for _ in range(n_ports):
+        for _ in range(n_ports // len(self.worker_nodes)):
+            tasks = [fun(self, 1, passive) for fun in self.worker_nodes]
+            L = await asyncio.gather(*tasks)
+            ret_list.append([x[0] for x in L])
+        
+        for _ in range(n_ports % len(self.worker_nodes)):
             worker = self.select_worker_for_port()
             ports = await worker.provision_ports(self, 1, passive)
             ret_list.append(ports[0])

putnopvut: Hi Mohammad. Thanks for the suggestion. I updated the PR with your suggestion. I added it on as a final commit. I did things a bit differently though. Rather than performing batches of provisioning based on the number of worker nodes, I instead grouped all of the provisioning into a single array and gathered that. It made the code a bit more readable IMO.

physical-deployments/physical-deployment.yml
```diff
@@ -1,9 +1,8 @@
-registry-node: host01.mydomain.com
+registry-node: wsfd-netdev72.ntdv.lab.eng.bos.redhat.com
```
mohammadheib, line 1: hi please revert this file :)

ovn-tester/ovn_workload.py
```diff
@@ -553,90 +574,97 @@ def __init__(self, central_node, worker_nodes, cluster_cfg, brex_cfg):
         self.last_selected_worker = 0
         self.n_ns = 0
 
-    def start(self):
-        self.central_node.start(self.cluster_cfg)
+    async def start(self):
+        await self.central_node.start(self.cluster_cfg)
         for w in self.worker_nodes:
```
mohammadheib, line 588: Hi @putnopvut, 
i think here we can start and config workers simultaneously as well 
i tried it with the following change and it's working fine:
```

    async def start(self):
         await self.central_node.start(self.cluster_cfg)
+        start_tasks = []
+        config_tasks = []
+        ctx = ovn_context.active_context
         for w in self.worker_nodes:
-            await w.start(self.cluster_cfg)
-            await w.configure(self.brex_cfg.physical_net)
+            start_tasks.append(ctx.create_task(w.start(self.cluster_cfg)))
+            config_tasks.append(ctx.create_task(w.configure(self.brex_cfg.physical_net)))
+        
+        await asyncio.gather(*start_tasks)
+        await asyncio.gather(*config_tasks2)
```

physical-deployments/physical-deployment.yml
putnopvut: Whoops! Thanks!

ovn-tester/ovn_workload.py
putnopvut: Excellent! I'll fold this change in. Is there a typo in the final line? Should that be `config_tasks` instead of `config_tasks2` ?

mohammadheib: yes, it is config_tasks :) sorry for the typo

ovn-tester/ovn_workload.py
putnopvut: Hi Mohammad. I implemented this suggestion, and then I tried it out on a test machine. Even though I can't see any issue with the code, it fails consistently for me.

I added some debug. With the original code, this is what I see happen:
```
2021-09-01 14:06:09,657 | ovn_workload |INFO| Creating localnet on ovn-scale-0
2021-09-01 14:06:09,657 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex"
2021-09-01 14:06:09,744 | ovn_workload |INFO| Adding external host on ovn-scale-0
2021-09-01 14:06:09,744 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1"
2021-09-01 14:06:09,838 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1"
2021-09-01 14:06:09,939 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1 exited with status 2
2021-09-01 14:06:09,939 | ovn_sandbox  |WARNING| stderr: RTNETLINK answers: File exists

2021-09-01 14:06:09,939 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 14:06:09,939 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip netns add ext-ns"
2021-09-01 14:06:10,039 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link set netns ext-ns dev veth0"
2021-09-01 14:06:10,142 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip netns exec ext-ns ip link set dev veth0 up"
2021-09-01 14:06:10,236 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip netns exec ext-ns ip addr add 3.0.255.253/16 dev veth0"
2021-09-01 14:06:10,339 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip netns exec ext-ns ip route add default via 3.0.255.254"
2021-09-01 14:06:10,422 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link set dev veth1 up"
2021-09-01 14:06:10,524 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ovs-vsctl add-port br-ex veth1"
```

The `ovn-scale-0` container is created, and a bunch of commands to set up network interfaces is run. Each of those succeeds (with the benign error of adding the same veth0 interface twice).

When I switch to using `asyncio.gather()`, it appears the worker container is not fully started up before we start issuing network commands. Adding some debug, I see results like this:
```
2021-09-01 13:52:32,877 | ovn_workload |INFO| Starting worker ovn-scale-0
2021-09-01 13:52:32,877 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "cd /root/ovn-heater/runtime/ovn-fake-multinode && OVN_MONITOR_ALL=yes OVN_DB_CLUSTER=yes ENABLE_SSL=yes ENABLE_ETCD=no OVN_DP_TYPE=system CREATE_FAKE_VMS=no CHASSIS_COUNT=0 GW_COUNT=0 RELAY_COUNT=3 IP_HOST=192.16.0.0 IP_CIDR=16 IP_START=192.16.0.2 ./ovn_cluster.sh add-chassis ovn-scale-0 tcp:0.0.0.1:6642"
2021-09-01 13:52:32,878 | ovn_workload |INFO| Creating localnet on ovn-scale-0
2021-09-01 13:52:32,878 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex"
2021-09-01 13:52:32,879 | ovn_workload |INFO| Starting worker ovn-scale-1
2021-09-01 13:52:32,879 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "cd /root/ovn-heater/runtime/ovn-fake-multinode && OVN_MONITOR_ALL=yes OVN_DB_CLUSTER=yes ENABLE_SSL=yes ENABLE_ETCD=no OVN_DP_TYPE=system CREATE_FAKE_VMS=no CHASSIS_COUNT=0 GW_COUNT=0 RELAY_COUNT=3 IP_HOST=192.16.0.0 IP_CIDR=16 IP_START=192.16.0.3 ./ovn_cluster.sh add-chassis ovn-scale-1 tcp:0.0.0.1:6642"
2021-09-01 13:52:32,880 | ovn_workload |INFO| Creating localnet on ovn-scale-1
2021-09-01 13:52:32,880 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-1 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex"
2021-09-01 13:52:32,938 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-1 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex exited with status 1
2021-09-01 13:52:32,938 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-1

2021-09-01 13:52:32,938 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:32,938 | ovn_workload |INFO| Adding external host on ovn-scale-1
2021-09-01 13:52:32,938 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1"
2021-09-01 13:52:32,939 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-0 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex exited with status 1
2021-09-01 13:52:32,939 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-0

2021-09-01 13:52:32,939 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:32,939 | ovn_workload |INFO| Adding external host on ovn-scale-0
2021-09-01 13:52:32,939 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1"
2021-09-01 13:52:32,984 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1 exited with status 1
2021-09-01 13:52:32,984 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-0

2021-09-01 13:52:32,984 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:32,984 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1"
2021-09-01 13:52:32,985 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1 exited with status 1
2021-09-01 13:52:32,985 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-1

2021-09-01 13:52:32,985 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:32,985 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1"
2021-09-01 13:52:33,031 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1 exited with status 1
2021-09-01 13:52:33,031 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-1

2021-09-01 13:52:33,032 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:33,032 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-1 ip netns add ext-ns"
2021-09-01 13:52:33,033 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1 exited with status 1
2021-09-01 13:52:33,033 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-0
...
```
As you can see, it now reports errors about the containers not existing. I'm guessing this is a temporary issue. By the time we move on in the tests and start trying to execute pings on those containers, we no longer get the "no such container" errors. However, since the network interfaces were not set up properly on those containers, it means that the pings fail.

I think we need a more consistent way of knowing the container in question is now up and running. We currently assume that if the SSH command returns, the container is ready. But that appears not to be the case.

As such, I'm going to revert this suggestion for the time being, pending a change for container readiness detection.

ovn-tester/ovn_workload.py
putnopvut: Hi Mohammad. I implemented this suggestion, and then I tried it out on a test machine. Even though I can't see any issue with the code, it fails consistently for me.

I added some debug. With the original code, this is what I see happen:
```
2021-09-01 14:06:09,657 | ovn_workload |INFO| Creating localnet on ovn-scale-0
2021-09-01 14:06:09,657 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex"
2021-09-01 14:06:09,744 | ovn_workload |INFO| Adding external host on ovn-scale-0
2021-09-01 14:06:09,744 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1"
2021-09-01 14:06:09,838 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1"
2021-09-01 14:06:09,939 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1 exited with status 2
2021-09-01 14:06:09,939 | ovn_sandbox  |WARNING| stderr: RTNETLINK answers: File exists

2021-09-01 14:06:09,939 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 14:06:09,939 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip netns add ext-ns"
2021-09-01 14:06:10,039 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link set netns ext-ns dev veth0"
2021-09-01 14:06:10,142 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip netns exec ext-ns ip link set dev veth0 up"
2021-09-01 14:06:10,236 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip netns exec ext-ns ip addr add 3.0.255.253/16 dev veth0"
2021-09-01 14:06:10,339 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip netns exec ext-ns ip route add default via 3.0.255.254"
2021-09-01 14:06:10,422 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link set dev veth1 up"
2021-09-01 14:06:10,524 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ovs-vsctl add-port br-ex veth1"
```

The `ovn-scale-0` container is created, and a bunch of commands to set up network interfaces is run. Each of those succeeds (with the benign error of adding the same veth0 interface twice).

When I switch to using `asyncio.gather()`, it appears the worker container is not fully started up before we start issuing network commands. Adding some debug, I see results like this:
```
2021-09-01 13:52:32,877 | ovn_workload |INFO| Starting worker ovn-scale-0
2021-09-01 13:52:32,877 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "cd /root/ovn-heater/runtime/ovn-fake-multinode && OVN_MONITOR_ALL=yes OVN_DB_CLUSTER=yes ENABLE_SSL=yes ENABLE_ETCD=no OVN_DP_TYPE=system CREATE_FAKE_VMS=no CHASSIS_COUNT=0 GW_COUNT=0 RELAY_COUNT=3 IP_HOST=192.16.0.0 IP_CIDR=16 IP_START=192.16.0.2 ./ovn_cluster.sh add-chassis ovn-scale-0 tcp:0.0.0.1:6642"
2021-09-01 13:52:32,878 | ovn_workload |INFO| Creating localnet on ovn-scale-0
2021-09-01 13:52:32,878 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex"
2021-09-01 13:52:32,879 | ovn_workload |INFO| Starting worker ovn-scale-1
2021-09-01 13:52:32,879 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "cd /root/ovn-heater/runtime/ovn-fake-multinode && OVN_MONITOR_ALL=yes OVN_DB_CLUSTER=yes ENABLE_SSL=yes ENABLE_ETCD=no OVN_DP_TYPE=system CREATE_FAKE_VMS=no CHASSIS_COUNT=0 GW_COUNT=0 RELAY_COUNT=3 IP_HOST=192.16.0.0 IP_CIDR=16 IP_START=192.16.0.3 ./ovn_cluster.sh add-chassis ovn-scale-1 tcp:0.0.0.1:6642"
2021-09-01 13:52:32,880 | ovn_workload |INFO| Creating localnet on ovn-scale-1
2021-09-01 13:52:32,880 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-1 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex"
2021-09-01 13:52:32,938 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-1 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex exited with status 1
2021-09-01 13:52:32,938 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-1

2021-09-01 13:52:32,938 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:32,938 | ovn_workload |INFO| Adding external host on ovn-scale-1
2021-09-01 13:52:32,938 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1"
2021-09-01 13:52:32,939 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-0 ovs-vsctl -- set open_vswitch . external-ids:ovn-bridge-mappings=providernet:br-ex exited with status 1
2021-09-01 13:52:32,939 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-0

2021-09-01 13:52:32,939 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:32,939 | ovn_workload |INFO| Adding external host on ovn-scale-0
2021-09-01 13:52:32,939 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1"
2021-09-01 13:52:32,984 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1 exited with status 1
2021-09-01 13:52:32,984 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-0

2021-09-01 13:52:32,984 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:32,984 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1"
2021-09-01 13:52:32,985 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1 exited with status 1
2021-09-01 13:52:32,985 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-1

2021-09-01 13:52:32,985 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:32,985 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1"
2021-09-01 13:52:33,031 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1 exited with status 1
2021-09-01 13:52:33,031 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-1

2021-09-01 13:52:33,032 | ovn_sandbox  |WARNING| stdout: 
2021-09-01 13:52:33,032 | ovn_sandbox  |INFO| Logging command: ssh wsfd-netdev72.ntdv.lab.eng.bos.redhat.com "docker exec ovn-scale-1 ip netns add ext-ns"
2021-09-01 13:52:33,033 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-0 ip link add veth0 type veth peer name veth1 exited with status 1
2021-09-01 13:52:33,033 | ovn_sandbox  |WARNING| stderr: Error: No such container: ovn-scale-0
...
```
As you can see, it now reports errors about the containers not existing. I'm guessing this is a temporary issue. By the time we move on in the tests and start trying to execute pings on those containers, we no longer get the "no such container" errors. However, since the network interfaces were not set up properly on those containers, it means that the pings fail.

I think we need a more consistent way of knowing the container in question is now up and running. We currently assume that if the SSH command returns, the container is ready. But that appears not to be the case.

As such, I'm going to revert this suggestion for the time being, pending a change for container readiness detection.

putnopvut: Hi Mohammad, I figured out the problem. The issue is that `ctx.create_task()` immediately starts the task. So if we create start tasks and configure tasks, they'll both start running ASAP. However, we need to ensure the start tasks complete before we even create the config tasks.

Something like this should work:
```
         await self.central_node.start(self.cluster_cfg)
-         for w in self.worker_nodes:
-            await w.start(self.cluster_cfg)
-            await w.configure(self.brex_cfg.physical_net) 
+        ctx = ovn_context.active_context
+       await asyncio.gather(*[ctx.create_task(w.start(self.cluster_cfg)) for w in self.worker_nodes])
+       await asyncio.gather(*[ctx.create_task(w.configure(self.brex_cfg.physical_net)) for w in self.worker_nodes])
```

mohammadheib: @putnopvut ohh got it,  great catch :) 
anyway, I think we still can add simultaneously provisioning to other parts of the code :) but for now, maybe you should merge the PRs and will push the other suggestions as a new PRs just to avoid breaking the code again :)

ovn-tester/ovn_sandbox.py
```diff
@@ -1,66 +1,86 @@
...
+            stdout.write(result.stdout.strip())
         else:
-            out = self.printable_result(ssh_stdout.read().decode().strip())
-            if len(out):
+            out = self.printable_result(result.stdout.strip())
+            if len(out) > 0:
                 log.info(out)
 
 
+async def create_physical_node(hostname, log_cmds):
```
mohammadheib, line 68: Hi @putnopvut,
i think here we only handled one ssh session per node and that could be a bottleneck for provisioning I'm not sure if that is something that could be changed but I tried it with multisession and seems that improve the provisioning please take a look on the change below :

ovn-tester/ovn_sandbox.py
```
@@ -1,4 +1,5 @@
 import logging
+import asyncio
 import asyncssh
 
 from ovn_exceptions import SSHError
@@ -12,6 +13,7 @@ asyncssh.set_log_level(logging.WARNING)
 
 async def create_ssh(hostname, log):
     ssh = SSH(hostname, log)
     # XXX Need to mirror the paramiko auto add policy?
     ssh.conn = await asyncssh.connect(hostname)
     return ssh
@@ -65,22 +67,30 @@ class SSH:
         return await self.conn.create_process(cmd)
 
 
-async def create_physical_node(hostname, log_cmds):
-    node = PhysicalNode()
-    node.ssh = await create_ssh(hostname, log_cmds)
+async def create_physical_node(hostname, log_cmds, num_conn=1):
+    node = PhysicalNode(num_conn)
+    for _ in range(node.num_conn):
+        node.ssh.append(await create_ssh(hostname, log_cmds))
     return node
 
 
 class PhysicalNode(object):
-    def __init__(self):
-        self.ssh = None
+    def __init__(self, num_conn=1):
+        self.ssh = []
+        self.num_conn = num_conn
+        self.ssh_sem = asyncio.Semaphore(self.num_conn)
+        self.free_sessions = list(range(self.num_conn))
 
     async def run(self, cmd="", stdout=None, raise_on_error=False):
-        await self.ssh.run(cmd=cmd, stdout=stdout,
+        async with self.ssh_sem:
+            # pop is atomic ops
+            session = self.free_sessions.pop()
+            await self.ssh[session].run(cmd=cmd, stdout=stdout,
                            raise_on_error=raise_on_error)
+            self.free_sessions.append(session)
 
     async def run_deferred(self, cmd=""):
-        return await self.ssh.run_deferred(cmd=cmd)
+        return await self.ssh[0].run_deferred(cmd=cmd)
 
 
```




ovn-tester/ovn_tester.py
```
@@ -91,7 +91,7 @@ async def read_physical_deployment(deployment, global_cfg):
 
         central_dep = dep['central-node']
         central_node = await create_physical_node(
-            central_dep.get('name', 'localhost'), global_cfg.log_cmds)
+            central_dep.get('name', 'localhost'), global_cfg.log_cmds, num_conn=25)
         worker_nodes = [
             await create_physical_node(worker, global_cfg.log_cmds)
             for worker in dep['worker-nodes']
@@ -239,7 +239,6 @@ async def main(global_cfg, cluster_cfg, brex_cfg, bringup_cfg):
     central, workers = await read_physical_deployment(sys.argv[1], global_cfg)
     central_node, worker_nodes = create_nodes(cluster_cfg, central, workers)
     tests = configure_tests(config, central_node, worker_nodes)
-
     ovn = await prepare_test(central_node, worker_nodes, cluster_cfg, brex_cfg)
     await run_base_cluster_bringup(ovn, bringup_cfg)
     for test in tests:
```

ovn-tester/ovn_workload.py
```diff
@@ -115,208 +120,218 @@ def __init__(self, phys_node, container, mgmt_net, mgmt_ip,
...
-                              logical_ip=cluster.net)
+        await cluster.nbctl.nat_add(self.gw_router, external_ip=str(gr_gw),
+                                    logical_ip=cluster.net)
 
     @ovn_stats.timeit
-    def provision_port(self, cluster, passive=False):
-        name = f'lp-{self.id}-{self.next_lport_index}'
-        ip = netaddr.IPAddress(self.int_net.first + self.next_lport_index + 1)
+    async def provision_port(self, cluster, passive=False):
+        async with self.lock:
```
dceara, line 261: As mentioned in the commit log that introduces the lock, this lock is not needed.  There's no `await` in the "critical section".  I would remove it.

ovn-tester/ovn_workload.py
```diff
@@ -350,206 +367,218 @@ def ping_ports(self, cluster, ports):
...
+        self.load_balancer = await create_load_balancer(f'lb_{self.name}',
+                                                        self.nbctl)
 
     @ovn_stats.timeit
-    def provision_vips_to_load_balancers(self, backend_lists):
-        vip_net = DEFAULT_NS_VIP_SUBNET.next(self.cluster.n_ns)
-        n_vips = len(self.load_balancer.vips.keys())
-        vip_ip = vip_net.ip.__add__(n_vips + 1)
+    async def provision_vips_to_load_balancers(self, backend_lists):
+        async with self.lock:
```
dceara, line 571: I think we can avoid this lock all together if we split the `add_vips` API into, e.g., `reserve_vips()` (synchronous and just adding the items to the load balancer vips dict and returning them) and `add_vips()` coroutine to actually set the vips.  If you agree, I'll let it up to you whether we should make the change in this PR or in a separate one.

ovn-tester/ovn_workload.py
```diff
@@ -567,91 +596,102 @@ def __init__(self, central_node, worker_nodes, cluster_cfg, brex_cfg):
...
-                [f'{p.ip}:{DEFAULT_BACKEND_PORT}' for p in ports]
-            for i, ports in enumerate(backend_lists)
-        }
-        self.load_balancer.add_vips(vips)
-
-    def unprovision_vips(self):
-        self.load_balancer.clear_vips()
-        self.load_balancer.add_vips(self.cluster_cfg.static_vips)
+    async def provision_vips_to_load_balancers(self, backend_lists):
+        async with self.lock:
```
dceara, line 675: Same comment as for `Namespace.provision_vips_to_load_balancers()`, I think we can avoid the lock.

ovn-tester/ovn_context.py
```diff
@@ -32,27 +38,75 @@ def __exit__(self, type, value, traceback):
...
+
+    async def qps_test(self, qps, coro, *args):
+        tasks = []
+        for i in range(self.max_iterations):
+            iteration = ContextIteration(i, self)
+            tasks.append(self.create_task(
+                self.qps_task(iteration, coro, *args), iteration)
+            )
+            if i % qps == 0:
+                await asyncio.sleep(1)
```
dceara, line 79: This is cool!  Although, if I'm reading this properly, super nit: `if i != 0 and i % qps == 0:`

dceara (commented):
This looks good overall, thanks!  I only left a few minor comments and questions.  Also, sorry, but you have to rebase.

ovn-tester/ovn_load_balancer.py
```diff
@@ -11,8 +11,21 @@ def __str__(self):
         return f"Invalid Protocol: {self.args}"
 
 
+async def create_load_balancer(lb_name, nbctl, vips=None,
```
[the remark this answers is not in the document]
putnopvut: Thanks for the suggestion. I thought about doing it this way. In the end, I went with the method in this PR because it seemed less likely to lead to errors.

For example, if we go with the suggested method, then if I wish to create a load balancer, I'd do something like this:
```
from ovn_load_balancer import OvnLoadBalancer
...
load_balancer = await OvnLoadBalancer.create(...)
```
However, because the entire `OvnLoadBalancer` class is imported, it would still be possible to do something like
```
load_balancer = OvnLoadBalancer(...)
```
This would result in a partially initialized load balancer being created and likely causing errors later. Best of luck to anyone trying to debug this :) With the method I used, it allows for this pattern:
```
from ovn_load_balancer import create_load_balancer
...
load_balancer = await create_load_balancer(...)
```
With this import, it is not possible to construct a load balancer directly.

There is nothing stopping people from importing the entire `ovn_load_balancer` file or the `OvnLoadBalancer` class with my method. I'm hopeful that the precedent set in this PR would be followed, though.

I'm willing to change over to using a `staticmethod` factory method if there are reasons for doing so that I'm not seeing. I figured I'd plead my case here first.

I really wish Python supported private methods.

ovn-tester/ovn_sandbox.py
putnopvut: Hi Mohammad, thanks for the suggestion. My main concern with this approach is the semaphore. If there are more than `num_conn` SSH commands occurring simultaneously, then tasks will have to wait for the semaphore to become available. As the scale increases, the waits for the semaphore could result in slower test runs. Worse, if we are timing the operation, then the time spent waiting for the semaphore would factor into the reported time.

On the other hand, my current code can result in a huge number of channels being created in a single SSH connection. That also has potential not to scale well.

What if we went with something like this instead (untested. There may be syntax errors):
```
-async def create_physical_node(hostname, log_cmds):
+async def create_physical_node(hostname, log_cmds, num_conn=1):
     node = PhysicalNode()
-    node.ssh = await create_ssh(hostname, log_cmds)
+    for _ in range(num_conn):
+        node.ssh.append(await create_ssh(hostname, log_cmds))
     return node
 
 
 class PhysicalNode(object):
     def __init__(self):
-        self.ssh = None
+        self.ssh = []
+        self.cur_conn = 0
 
     async def run(self, cmd="", stdout=None, raise_on_error=False):
-        await self.ssh.run(cmd=cmd, stdout=stdout,
-                           raise_on_error=raise_on_error)
+        self.cur_conn += 1
+        if self.cur_conn >= len(self.ssh):
+            self.cur_conn = 0
+
+        await self.ssh[self.cur_conn].run(cmd=cmd, stdout=stdout,
+                                          raise_on_error=raise_on_error)
 
     async def run_deferred(self, cmd=""):
-        return await self.ssh.run_deferred(cmd=cmd)
+        return await self.ssh[0].run_deferred(cmd=cmd)
```
This uses your suggestion of having a pool of SSH connections. However, instead of performing at most one command per connection, we cycle through the connections round-robin. This makes it more likely that each connection is only using a small to medium number of channels. This also ensures that we will never wait before attempting to send an SSH command. What do you think?

ovn-tester/ovn_workload.py
putnopvut: I was messing with this just now, and removing the locking is going to be non-trivial. 

Adding VIPs is not an additive operation. Under the hood it calls `ovn-nbctl set Load_Balancer vips ...`. That means each individual task is re-setting the entire VIP list instead of just adding their own VIPs to the configured load balancer. This means that each task needs to have exclusive access to the load balancer.

The VIPs are added in batches, meaning that each time we add new VIPs, we might end up making multiple ovn-nbctl calls per load balancer. Since all of those ovn-nbctl calls are asynchronous, it gives other tasks the opportunity to foul up the configured VIPs. So even if we could make the locking finer-grained, I think we'd still need locking as long as we're batching the VIPs.

Having said that, I actually noticed that there's a bug in the current `add_vips()` method due to the batching code. Since we are setting the VIPs on the load balancer instead of adding to them, it means that each batch undoes the previous batch's operation. Effectively it means that a load balancer will never have more than `MAX_VIPS_IN_BATCH` (currently 500) vips configured. This may not be as big an issue with load balancer per service, but it's still a bug. Since the operation is not additive, I actually have no idea how batching can work here.

If we could make adding VIPs additive, then this would be easy since each task would be performing independent non-overlapping operations. It would also allow for proper batching.

ovn-tester/ovn_load_balancer.py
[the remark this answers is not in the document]
dceara: It's ok, thanks for the rationale, let's leave it as is.

ovn-tester/ovn_sandbox.py
mohammadheib: @putnopvut Hi Mark,
yes your suggestions looks very good to me :+1:

ovn-tester/ovn_workload.py
dceara: > I was messing with this just now, and removing the locking is going to be non-trivial.
> 
> Adding VIPs is not an additive operation. Under the hood it calls `ovn-nbctl set Load_Balancer vips ...`. That means each individual task is re-setting the entire VIP list instead of just adding their own VIPs to the configured load balancer. This means that each task needs to have exclusive access to the load balancer.

Actually, I think adding VIPs *is* an additive operation, it translates to something equivalent to:

```
$ ovn-nbctl lb-add lb-test 42.42.42.42:4242 21.21.21.21:2121
$ ovn-nbctl list load_balancer
_uuid               : 00874eb0-2266-43da-870c-3e8f9cb55617
name                : lb-test
vips                : {"42.42.42.42:4242"="21.21.21.21:2121"}
$ ovn-nbctl set load_balancer lb-test vips:\"43.43.43.43:4343\"=\"22.22.22.22:222\"
$ ovn-nbctl list load_balancer
_uuid               : 00874eb0-2266-43da-870c-3e8f9cb55617
name                : lb-test
vips                : {"42.42.42.42:4242"="21.21.21.21:2121", "43.43.43.43:4343"="22.22.22.22:222"
```

> 
> The VIPs are added in batches, meaning that each time we add new VIPs, we might end up making multiple ovn-nbctl calls per load balancer. Since all of those ovn-nbctl calls are asynchronous, it gives other tasks the opportunity to foul up the configured VIPs. So even if we could make the locking finer-grained, I think we'd still need locking as long as we're batching the VIPs.
> 
> Having said that, I actually noticed that there's a bug in the current `add_vips()` method due to the batching code. Since we are setting the VIPs on the load balancer instead of adding to them, it means that each batch undoes the previous batch's operation. Effectively it means that a load balancer will never have more than `MAX_VIPS_IN_BATCH` (currently 500) vips configured. This may not be as big an issue with load balancer per service, but it's still a bug. Since the operation is not additive, I actually have no idea how batching can work here.

As mentioned above, unless I'm missing something, the operation is already additive.

> 
> If we could make adding VIPs additive, then this would be easy since each task would be performing independent non-overlapping operations. It would also allow for proper batching.

putnopvut: > Actually, I think adding VIPs **is** an additive operation, it translates to something equivalent to:
> 
> ```
> $ ovn-nbctl lb-add lb-test 42.42.42.42:4242 21.21.21.21:2121
> $ ovn-nbctl list load_balancer
> _uuid               : 00874eb0-2266-43da-870c-3e8f9cb55617
> name                : lb-test
> vips                : {"42.42.42.42:4242"="21.21.21.21:2121"}
> $ ovn-nbctl set load_balancer lb-test vips:\"43.43.43.43:4343\"=\"22.22.22.22:222\"
> $ ovn-nbctl list load_balancer
> _uuid               : 00874eb0-2266-43da-870c-3e8f9cb55617
> name                : lb-test
> vips                : {"42.42.42.42:4242"="21.21.21.21:2121", "43.43.43.43:4343"="22.22.22.22:222"
> ```

That works as long as the key for the VIP is different each time. It appears our tests do not attempt to add backends to existing VIPs, so you're right this should work for our purposes.

If we can guarantee that the set of keys in the dictionary of VIPs passed to `add_vips()` does not intersect with the set of keys already configured on the load balancer, we could also simplify some of the code in `add_vips()`

dceara: Shall we raise an exception then if a test tries to add backends to an already configured VIP?

ovn-tester/ovn_workload.py
```diff
@@ -649,12 +649,14 @@ async def create_cluster_join_switch(self, sw_name):
         )
 
     async def provision_ports(self, n_ports, passive=False):
-        ret_list = []
+        tasks = []
+        ctx = ovn_context.active_context
```
mohammadheib, line 653: Hi, 
i think here we can save few seconds by looping over workers instead of ports and create a bundle of ports per ports for each loop iteration i'm not sure if that could really improve the timing of the tests but i saw some improvement on my tests:
```

@@ -651,10 +659,15 @@ class Cluster(object):
     async def provision_ports(self, n_ports, passive=False):
         tasks = []
         ctx = ovn_context.active_context
-        for _ in range(n_ports):
+        
+        for _ in range(len(self.worker_nodes)):
             worker = self.select_worker_for_port()
-            tasks.append(ctx.create_task(worker.provision_ports(self, 1,
+            tasks.append(ctx.create_task(worker.provision_ports(self, (n_ports//len(self.worker_nodes)),
                                                                 passive)))
+        for _ in range(n_ports%len(self.worker_nodes)):
+            worker = self.select_worker_for_port()
+            tasks.append(ctx.create_task(worker.provision_ports(self, 1, passive)))
+
         results = await asyncio.gather(*tasks)
         return [ports[0] for ports in results]
```

ovn-tester/ovn_context.py
putnopvut: I went a slightly different route:
```
if (i  + 1) % qps == 0:
    await asyncio.sleep(1)
```
The original version would sleep after iteration 0. Your suggestion fixes that problem, but it causes the first sleep to be performed after 21 iterations instead of 20.

putnopvut:
Hi, just leaving a comment to explain a few things. I have a couple more commits I want to add to this series and then I'll update the PR. I'm testing them in my `async_sandbox` branch right now. The first is one that changes how we wait for ports to be up. The second adds the connection pool as has been discussed in this PR.

When reviewing, I suspect you'll be able to find things that could be optimized, or you could find places where we can use asyncio features in ways that will be more efficient. But I think those sorts of improvements can wait until after the initial asyncio support is merged.

I also opened two issues: #85 and #86 . These issues exist without asyncio, but with asyncio they're even worse. I don't think those issues should impede us from getting the initial asyncio code in, though. These can be addressed afterwards.

It's currently Friday afternoon where I am so I've run out of time for the week. However, I plan on getting this PR updated early next week so that this can be reviewed again.

dceara (commented):
If I understand correctly, with this, the only (mostly) reliable measurement is the `wait_for_port_up` value because we use `ovn-installed-ts` to determine when a port is up (I'm saying "mostly" because the start timestamp is inaccurate, it's after `lsp-add` **and** `lsp-get` were executed).

Should we emphasize this in the report somehow? Or shall we remove the other measurements?

Also, do you have some test run results on larger scale deployments (more physical machines)?

PR #68, opened 2021-08-19
Set gateway chassis for each cluster router port.

numansiddique:
OVN supports creating multiple gw router ports for a logical router.
This support was added to improve the scalability.  Make use of it.
A gateway chassis is created for each router port of the cluster router
with the node chassis as gateway chassis.

Signed-off-by: Numan Siddique <nusiddiq@redhat.com>


numansiddique:
The corresponding ovn-k8s PR is merged.

I guess this PR can be considered now.

putnopvut:
Thanks Numan, looks good to me.

PR #70, opened 2021-08-30
do.sh: Collect data about poll intervals and database compactions.

igsilya:
This commit adds a couple of scripts that mines data about unreasonably
long poll intervals from logs and stores it in a human-readable format
with all the statistics including min, max, avg, 95 percentile, etc.
values per OVN daemon.  For databases it tries to exclude poll
intervals that involved compactions and also collects data about time
spent on compactions separately.  Logic for this is not ideal, but it
seems to work fine for most cases.

Statistics counted with 'datamash' utility which is a new dependency.

Example of the output:
```
  Unreasonably long poll intervals:

  ---------------------------------------------------------------------
      Count       Min        Max       Median      Mean   95 percentile
  ---------------------------------------------------------------------
      1073.0     1000.0     9341.0     1612.0     3483.7     8401.2
      2031.0     1011.0     9961.0     1544.0     2533.1     8157.0
      3332.0     1000.0     9501.0     1467.5     2698.3     8293.0
  ---------------------------------------------------------------------
      6436.0     1000.0     9961.0     1565.5     2777.1     8302.2
```

Here is a result per log file and the result when all log files
considered all together, i.e. 'Mean' on the last line is not an average
of 'Mean' values of all rows, but a 'Mean' value of the data from all
the files.

### DIFF
--- a/utils/mine-db-poll-intervals.sh
+++ b/utils/mine-db-poll-intervals.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+function mine_data_for_poll_intervals {
+    res=$(echo "$1" \
+        | sed '/\(writes\|involuntary\)/s/$/\n\n/' \
+        | egrep -A 2 -B 1 'Unreasonably' \
+        | grep -v '\-\-' | cut -c 45- | paste -d " " - - - - \
+        | sed 's/.*took \([0-9]*\)ms .*long \([0-9]*\)ms .* \([0-9]*\)ms system.* \([0-9]*\) writes.*/\2 \3 \4 compaction \1/' \
+        | sed 's/.*long \([0-9]*\)ms .* \([0-9]*\)ms system.*/\1 \2/' \
+        | sort -h | grep -v compaction | cut -f 1 -d ' ' \
+        | datamash count 1 min 1 max 1 median 1 mean 1 perc 1 \
+                   -t ' '  --format="%10.1f" )
+    if [ -z "$res" ]; then
+        printf "%10.1f\n" 0
+    else
+        echo "$res"
+    fi
+}
+
+function mine_data_for_compaction {
+    res=$(echo "$1" \
+        | grep 'compaction took' | cut -c 43- \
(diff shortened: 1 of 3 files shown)


utils/mine-db-poll-intervals.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+function mine_data_for_poll_intervals {
+    res=$(echo "$1" \
+        | sed '/\(writes\|involuntary\)/s/$/\n\n/' \
+        | egrep -A 2 -B 1 'Unreasonably' \
+        | grep -v '\-\-' | cut -c 45- | paste -d " " - - - - \
+        | sed 's/.*took \([0-9]*\)ms .*long \([0-9]*\)ms .* \([0-9]*\)ms system.* \([0-9]*\) writes.*/\2 \3 \4 compaction \1/' \
+        | sed 's/.*long \([0-9]*\)ms .* \([0-9]*\)ms system.*/\1 \2/' \
+        | sort -h | grep -v compaction | less | cut -f 1 -d ' ' \
```
dceara, line 10: Nit: `less` not needed.

Nit2: To some extent this is similar to `mine_data_for_poll_intervals()` from `mine-poll-intervals.sh`; I'd prefer not duplicating this complex data manipulation but if there's no easy way let's keep it as is for now.

igsilya: I'll remove `less`, thanks!  Construction of this pipeline was kind of iterative, so it seems like a leftover.  This function especially is a bit magical. :)
For the similarity, I agree that these functions duplicates each other, but I'm not sure what is a good way to break them into reusable modules.  The most I can do is to split the `datamash` part + check for zero out of them, I guess.  But, maybe, it's better to just keep them as is.  If we'll need more of these functions in the future,we can refactor that to a library of shell scripts later.

dceara: Ok, let's keep them as is for now.

dceara (approved):

PR #74, opened 2021-09-01
Batch load_balancer bind to switch/router operations.

dceara:
On scaled setups, since we moved to load balancer per namespace there
will be lots of load balancers that need to be bound to lots of node
gateway routers/switches.  Binding to one node at a time is just a waste
of time.  Change load balancer APIs to allow binding to multiple
routers/switches in one operation.

For example, on a 20 node cluster density test run (10 startup
iterations) this reduces the test runtime by ~4 minutes (~40%).


putnopvut (approved):
Looks good to me!

PR #75, opened 2021-09-01
ovn_workload: Always specify load balancer VIP ports.

dceara:
This is to better simulate what ovn-kubernetes does.


dceara:
CC: @igsilya

putnopvut (approved):

PR #76, opened 2021-09-01
ovn-tester: Don't try creating the veth twice.

putnopvut:
For some reason the exact same veth creation command is run twice in a
row when provisioning. Adding some debugging, I saw the following:

2021-09-01 14:06:15,391 | ovn_sandbox  |WARNING| Command docker exec ovn-scale-1 ip link add veth0 type veth peer name veth1 exited with status 2
2021-09-01 14:06:15,391 | ovn_sandbox  |WARNING| stderr: RTNETLINK answers: File exists

The error is benign, but we may as well get rid of it.


dceara:
Looks good to me, thanks!

PR #77, opened 2021-09-01
ovn-tester: Separate relays from other DBs.

putnopvut:
relay containers only run a southbound database. Therefore, when we try
to enable trim on compaction for the northbound database in that
container, it results in an error. The error is benign, but it still is
better not to issue a useless command.

This commit fixes the problem by separating relay containers from other
DB containers. This way, they can get special treatment if necessary. In
this particular case, the special treatment is not to try to configure a
non-existent database.

### DIFF
--- a/ovn-tester/ovn_tester.py
+++ b/ovn-tester/ovn_tester.py
@@ -193,9 +193,10 @@ def create_nodes(cluster_config, central, workers):
     ] if cluster_config.clustered_db else [
         'ovn-central'
     ]
-    for i in range(cluster_config.n_relays):
-        db_containers.append(f'ovn-relay-{i + 1}')
-    central_node = CentralNode(central, db_containers, mgmt_net, mgmt_ip)
+    relay_containers = [f'ovn-relay-{i + 1}' for i in
+                        range(cluster_config.n_relays)]
+    central_node = CentralNode(central, db_containers, relay_containers,
+                               mgmt_net, mgmt_ip)
     worker_nodes = [
         WorkerNode(workers[i % len(workers)], f'ovn-scale-{i}',
                    mgmt_net, mgmt_ip + i + 1, internal_net.next(i),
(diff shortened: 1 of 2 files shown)


ovn-tester/ovn_tester.py
```diff
@@ -193,9 +193,11 @@ def create_nodes(cluster_config, central, workers):
     ] if cluster_config.clustered_db else [
         'ovn-central'
     ]
+    relay_containers = []
     for i in range(cluster_config.n_relays):
-        db_containers.append(f'ovn-relay-{i + 1}')
-    central_node = CentralNode(central, db_containers, mgmt_net, mgmt_ip)
+        relay_containers.append(f'ovn-relay-{i + 1}')
```
dceara, line 198: Nit: wouldn't it be a bit more "pythonic" as below?
```
relay_containers = [f'ovn-relay-{i + 1}' for i in range(cluster_config.n_relays)]
```

ovn-tester/ovn_workload.py
```diff
@@ -99,6 +101,10 @@ def enable_trim_on_compaction(self):
             self.phys_node.run(f'docker exec {db_container} ovs-appctl -t '
                                f'/run/ovn/ovnsb_db.ctl '
                                f'ovsdb-server/memory-trim-on-compaction on')
+        for relay_container in self.relay_containers:
+            self.phys_node.run(f'docker exec {db_container} ovs-appctl -t '
```
dceara, line 105: This should be `relay_container` instead of `db_container`.

dceara (changes requested):
Just a tiny nit, if you fix it, please go ahead and merge

ovn-tester/ovn_workload.py
putnopvut: Whoops, this is fixed now.

ovn-tester/ovn_tester.py
putnopvut: I changed this based on your suggestion.

dceara (approved):
Looks good to me, thanks!

PR #78, opened 2021-09-01
ovn-tester: Unprovision build_ports via the namespace.

putnopvut:
In the cluster_density test, we create a list of build ports, followed
by other ports during an iteration. At the end of the iteration, we
unprovision the build ports.

If global configuration is set to clean up at the end of a test, we then
try to unprovision all ports in the namespace once the test completes.
The problem is, we already unprovisioned the build ports during each
iteration. The namespace doesn't know this, and it results in errors.

This commit adds a new method for Namespace to unprovision a subset of
ports. We now unprovision the build ports in the cluster_density test
using this method. Now when the test ends and the namespace is
unprovisioned, it only unprovisions the remaining ports.

### DIFF
--- a/ovn-tester/ovn_workload.py
+++ b/ovn-tester/ovn_workload.py
@@ -395,6 +395,15 @@ class Namespace(object):
         for addr_set in self.sub_as:
             self.nbctl.address_set_del(addr_set)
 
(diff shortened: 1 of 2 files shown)


dceara:
Looks good to me, thanks!

PR #84, opened 2021-10-01
density_heavy: Fix startup stage.

dceara:
We were incorrectly provisioning way too many backends for every load
balancer created in the "startup" stage.  We should instead provision
one backend every 'pods_vip_ratio' pods in 'n_startup'.


dceara:
CC: @LorenzoBianconi @putnopvut, I'm merging this directly because it blocks our downstream CI and it's a straightforward fix.  Please let me know if you think a follow up is needed.

PR #91, opened 2021-12-13
Add load balancer groups.

dceara:
This applies to all tests that use Cluster.provision_lb(), which today
corresponds to density-heavy and cluster-density.


dceara:
As discussed offline with @LorenzoBianconi this works fine, merging.

PR #93, opened 2022-01-10
Use UUIDs for records not indexed by name.

dceara:


igsilya:
I didn't test this but looks sane. :)

dceara:
> I didn't test this but looks sane. :)

Thanks!

PR #100, opened 2022-04-08
do.sh: Fail installation if building the container image fails.

dceara:


dceara:
For a successful installation and test run we need the ovn-fake-multinode [PR to build OVN against the OVS submodule](https://github.com/ovn-org/ovn-fake-multinode/pull/56) to be merged first.

almusil (approved):

PR #104, opened 2022-06-02
Reformat ovn-heater with Black

almusil:
Reformat the codebase with Black
and add it to CI runs. To format
the files locally you can run:
black -l 79 -S $FILE

Signed-off-by: Ales Musil <amusil@redhat.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ on:
     branches: [ main ]
 
 jobs:
-  build:
+  lint:
 
     runs-on: ubuntu-latest
     strategy:
@@ -23,7 +23,11 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install flake8
+        python -m pip install flake8 black==22.3.0
+    # Ignore E203, W503 and W504 which is against PEP 8 style
     - name: flake8
       run: |
-        flake8 ovn-tester
+        flake8 --ignore=E203,W504,W503 ovn-tester
+    - name: Black
+      run: |
+        black -l 79 -S --check --diff . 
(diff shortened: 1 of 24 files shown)


LorenzoBianconi:
@almusil I am fine with this change, but I think we should postpone it after IPv6 support

almusil:
Definitely, that's why it is draft, I don't want to cause rebase pain :)

dceara:
@almusil the IPv6 support is merged.  Do you plan to rebase this PR and turn it into a formal one?

putnopvut:
+1 on getting this merged soon. I tend to have my python files auto-formatted with Black when I save them. I currently have to disable that with ovn-heater since it results in huge diffs otherwise.

almusil:
Rebased

.github/workflows/test.yml
```diff
@@ -23,7 +23,11 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install flake8
+        python -m pip install flake8 black==22.3.0
+    # Ignore E203, W503 and W504 which is against PEP 8 style
     - name: flake8
       run: |
-        flake8 ovn-tester
+        flake8 --ignore=E203,W504,W503 .
```
dceara, line 30: This now runs flake8 on more files which is a good thing.  But we should also fix the warnings it raises in those files, e.g.:
```
./ovn-fake-multinode-utils/process-monitor.py:51:9: E722 do not use bare 'except'
./utils/process-stats.py:2:1: F401 'numpy' imported but unused
./utils/process-stats.py:7:1: F401 'time' imported but unused
```

almusil: Ah didn't see those as the CI was blocked, ack I will address it

dceara (approved):
Looks good to me, thanks!

PR #105, opened 2022-06-13
ovn_workload: Log creation of load balancers.

igsilya:
ocp-120-density-heavy scenario creates 5000 load balancers
during the "startup" phase.  It takes 3-4 hours and there
are no logs printed during that time.

Fix that by actually logging creation of load balancers.


dceara:
Looks good to me, thanks!

PR #106, opened 2022-06-13
Collect RSS and CPU usage for OVS/OVN processes.

igsilya:
With this change the new process-monitor.py script is copied into
each container and started to collect RSS and CPU usage of all
OVS/OVN processes.  Later, log-collector will stop the monitoring
and will collect results as a process-stats.json file.  do.sh
script then uses process-stats.py script at the data mining stage
to produce html reports for master and worker nodes containing
RSS an CPU usage graphs.

The integration part is a bit sketchy.  Alternative is to avoid
modification of the ovn-tester itself and only deploy the
process monitor on the physical host.  This will make the code
a bit cleaner, but we'll loose an easy way to add container names
to process names.  Also, the monitor spends 0.5 seconds per process
to estimate the CPU usage.  More processes to track - less accurate
the results.  For example, one round of measurements for 9 processes
takes ~4.5 seconds, so we can easily miss the 3 second database
compaction.  We could run a monitor per process though, but that
sounds like an overkill.

In order to work properly, python3-psutil needs to be installed
in the container.  If it's not available, ovn-heater will still
work normally, but resource usage reports will not be generated.

### DIFF
--- a/do.sh
+++ b/do.sh
@@ -23,6 +23,7 @@ docker_daemon_file=${rundir}/docker-daemon.json
 podman_registry_file=${rundir}/registries.conf
 log_collector_file=${rundir}/log-collector.sh
 log_perf_file=${rundir}/perf.sh
+process_monitor_file=${rundir}/process-monitor.py
 
 ovn_tester=${topdir}/ovn-tester
 ovn_tester_log_file=test-log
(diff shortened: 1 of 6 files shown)


igsilya:
Corresponding ovn-fake-multinode PR: https://github.com/ovn-org/ovn-fake-multinode/pull/64

igsilya:
Reports look like this, but they are interactive html (github doesn't allow to attach html):
rss

flavio-fernandes:
Request: Make it an ovscon 2022 track!

dceara:
Looks great to me, thanks!

PR #107, opened 2022-06-21
do.sh: Gracefully deal with grep pipefail when processing stats.

dceara:
The 'head -3' command might close the pipe while grep is still
writing to it, causing a SIGPIPE, which in turn causes the script to
fail because we run 'do.sh' with 'pipefail' set.


dceara:
CC: @igsilya

PR #108, opened 2022-06-21
do.sh: Handle all expected pipe failures properly.

dceara:
Commit 8208181f1e33 ("do.sh: Gracefully deal with grep pipefail when processing
stats.") is incomplete in dealing with pipe failures because it only
took care of the first failure in the chain.  Properly deal with it by
adding 'or TRUE' at the end.


igsilya (approved):
Looks OK to me.  Thanks for the follow up!

PR #110, opened 2022-08-03
ovn-tester: Add 60 and 180 node OCP like test scenarios.

dceara:
Running tests at 20, 60, 120, 180, 250 node scales should help detect
whether OVN performance scales linearly or not.

### DIFF
--- a/test-scenarios/ocp-180-cluster-density.yml
+++ b/test-scenarios/ocp-180-cluster-density.yml
@@ -1,0 +1,14 @@
+global:
+  log_cmds: False
+
+cluster:
+  clustered_db: true
+  monitor_all: true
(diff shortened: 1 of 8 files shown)


igsilya (approved):
I see you chose 45 pods per node for all heavy tests.  LGTM.  We just need to keep in mind that 20 node test is different.

dceara:
> I see you chose 45 pods per node for all heavy tests. LGTM. We just need to keep in mind that 20 node test is different.

Yes, the 20 node heavy test matches the OCP scale test.  I'm not sure why they chose different densities for more nodes though.  In any case, you're right, when analyzing results we should take density into account too.

Thanks for the review!

PR #111, opened 2022-08-03
ovn-tester: Add missing whitespace after assert.

dceara:


dceara:
Checks passed, merging.

PR #112, opened 2022-08-09
Open Stack Scenario

ashwins108:


do.sh
```diff
@@ -289,7 +289,7 @@ function run_test() {
     source ${rundir}/${ovn_heater_venv}/bin/activate
     pushd ${out_dir}
 
-    if ! python -u ${ovn_tester}/ovn_tester.py $phys_deployment ${test_file} 2>&1 | tee ${ovn_tester_log_file}; then
```
dceara, line 292: This doesn't seem ok to me.  Moreover, why do we need a new ovn_opstack_tester.py?  Can't we include the functionality in the current ovn_tester.py?

dceara (changes requested):
Thanks for the contribution!  It's not really clear to me what type of scenario you're trying to implement.  Also, wouldn't it make sense to just implement it as a test scenario in `ovn-tester/tests`?

PR #113, opened 2022-08-10
ovn_workload: Log binding of a logical port.

igsilya:
Binding is one of the most heavy operations and it's not logged at the
moment.  We may not see any logs for a log time while ovn-heater is
binding thousands of ports during the test startup stage, for example.


dceara (approved):
Looks good to me, thanks!

PR #114, opened 2022-08-10
do.sh: Fix initial installation.

igsilya:
If there are no ovn/registry containers on a system, lookup will fail
and the 'pipefail' will fail the entire run.

If the registry container is stopped, it won't show up in the output
of a normal 'docker ps', but the subsequent 'docker run' will fail
to start container with the same name, so '--all' has to be used.

Also, if due to earlier failures everything is already checked out,
but the image doesn't exist, it will not be built.  So, we need to
check if the image is available and re-build if it is not.


dceara:
Looks good to me, thanks!

PR #115, opened 2022-08-11
do.sh: Add lport binding to ovn-installed latency to mined data.

igsilya:
Difference per logical port between port being bound to the node
and ovn-controller marking it as ovn-installed, including usual
min, max, avg, 95%ile.  Ports that was never actually installed
are also counted.

Results are in the mined-data/binding-to-ovn-installed-latency.

The metric should be taken with a grain of salt, because ports
are bound (OVS port with iface-id created) after they were already
created in Nb DB.  And for bulk port creations, binding happens
in bulk as well afterward.  The value can be more or less treated
as an end to end OVN latency for ports created during the test
phase though.

Useful to spot issues like very high latency or ports not being
installed at all.

ovn-heater might be changed in the future to always bind ports
right after creation to make the metric more useful.

### DIFF
--- a/do.sh
+++ b/do.sh
@@ -312,6 +312,19 @@ function run_test() {
         ${topdir}/utils/mine-db-poll-intervals.sh ${logs} > mined-data/${p}
     done
 
+    cat ${out_dir}/test-log | grep 'Binding lport' \
+        | cut -d ' ' -f 1,2,8 > mined-data/ovn-binding.log
+
+    logs=$(find ${out_dir}/logs -name ovn-controller.log)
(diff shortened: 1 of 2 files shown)


igsilya:
Output example:
```
min     : 13 ms
max     : 15665 ms
avg     : 1149 ms
95%     : 1711 ms
total   : 13750
failures: 0

lp-0-0    : 16
lp-0-1    : 14
lp-0-2    : 13
lp-0-3    : 13
...
```

dceara:
Nice!

PR #116, opened 2022-08-11
sandbox: Speed up command execution by re-using docker exec shell.

igsilya:
Currently ovn-heater calls 'docker exec' via SSH connection to a
physical host every time it needs to execute any command inside
a container.  Just 'docker exec' itself adds anywhere from 150 to
500 ms of extra overhead to every call.

To avoid that overhead, creating a separate channel for each
container within a common SSH connection.  Using an interactive
shell emulator from the invoke_shell() method, invoking a docker
interactive shell.  Once this is done, we have a direct channel
to execute any commands inside of a container with only few
milliseconds of overhead (4-5 ms in practice).

Since it's not just a command execution API like paramiko's
exec_command(), but a real shell emulator, it's a bit trickier
to work with.  Shell sends back command prompts and everything
you send to it.  It also sends back lots of vt100 control/escape
symbols.  Command output is filtered out by adding special
start/end markers, the exit code is printed with echo and
parsed out as well.

There are no functional changes to tests, they are doing exactly
the same to what they were doing before, but with the command
execution being much faster.

One consequence of commands being executed faster is added sleep
to the election timer modifications that should wait for cluster
to agree on a new value before it can be changed again.

This change greatly improves the test speed.  For example, 250
node density heavy test can be finished in 1 hour and 20 minutes
instead of 9 hours before.

### DIFF
--- a/ovn-tester/ovn_sandbox.py
+++ b/ovn-tester/ovn_sandbox.py
@@ -1,6 +1,8 @@
 import logging
 import paramiko
+import socket
 
+from io import StringIO
 from ovn_exceptions import SSHError
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


dceara:
Looks good to me, thanks!

PR #118, opened 2022-08-18
ovn-tester: Wait for all changes to be applied before exiting context.

igsilya:
Adding an extra 'ovn-nbctl sync' while exiting test contexts,
so previous test stages doesn't skew results for later ones.


dceara (approved):
Looks good to me, thanks!

PR #120, opened 2022-08-19
ovn_utils: Remove no longer necessary 'bash -c'.

dceara:


igsilya (approved):
As long as you tested it, LGTM.

I wonder if we should '&&' them, instead of ';', but that's a separate issue.

putnopvut:
I agree that switching to `&&` instead of `;` should be saved for a different enhancement.

PR #121, opened 2022-08-22
Update maximum number of SSH sessions and fs notifiers

igsilya:
Without these changes the host will not be able to run more than
10-ish containers.  The default value for MaxSessions is only 10,
defualt value for fs.inotify.max_user_instances also allows to
start only 10-12 containers on the host:

 ovn-scale-59: Failed to create control group inotify object: Too many open files
 ovn-scale-59: Failed to allocate manager object: Too many open files
 ovn-scale-59: [!!!!!!] Failed to allocate manager object.

Issues can be easily triggered by attempting to run 20 or 60
node tests on localhost.

Chosen values should allow to run any reasonable number of fake
nodes per physical machine.

### DIFF
--- a/ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
+++ b/ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
@@ -20,6 +20,16 @@
       ./ovn_cluster.sh stop
       docker system prune -f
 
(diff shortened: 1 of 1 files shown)


putnopvut:
I can confirm that on a Fedora 36 machine, this patch allowed me to have more than 20 containers running at once on a server.

PR #122, opened 2022-08-23
ovn_tester: Enable northd parallelization.

igsilya:
OpenShift is using northd parallelization with 4 threads by default.
ovn-heater should do that as well.  Can be configured by setting
'northd_threads' in the test yml.


putnopvut (approved):
Looks good to me. Great idea, Ilya.

dceara (approved):
Thanks!

PR #124, opened 2022-08-25
Ensure the neighbor cache size on central nodes is sufficient

igsilya:
In high-scale tests the kernel on central nodes complains:

  kernel: neighbour: arp_cache: neighbor table overflow!

This is due to large number of incoming database connections from
all other nodes.  Setting up higher cache size to avoid the problem.

Values are taken from a live ovn-kubernetes cluster.

Not using ansible's 'sysctl' module to avoid setting these values
permanently.

### DIFF
--- a/ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
+++ b/ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
@@ -30,6 +30,13 @@
       echo "MaxSessions 1024" >> /etc/ssh/sshd_config
       systemctl reload sshd.service
 
+  - name: Ensure the neighbor cache size on central nodes
+    when: ovn_central is defined
+    shell: |
+      sysctl -w net.ipv4.neigh.default.gc_thresh1=8192
+      sysctl -w net.ipv4.neigh.default.gc_thresh2=32768
+      sysctl -w net.ipv4.neigh.default.gc_thresh3=65536
+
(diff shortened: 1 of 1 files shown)


ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
```diff
@@ -30,6 +30,13 @@
       echo "MaxSessions 1024" >> /etc/ssh/sshd_config
       systemctl reload sshd.service
 
+  - name: Ensure the neighbor cache size on central nodes
+    when: ovn_central is defined
+    shell: |
+      sysctl -w net.ipv4.neigh.default.gc_thresh1 = 8192
```
dceara, line 36: This needs to be (without extra spaces):
```
      sysctl -w net.ipv4.neigh.default.gc_thresh1=8192
      sysctl -w net.ipv4.neigh.default.gc_thresh2=32768
      sysctl -w net.ipv4.neigh.default.gc_thresh3=65536
```

dceara (changes requested):

ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
igsilya: Oops.  Fixed.

dceara (approved):

PR #128, opened 2022-08-26
cirrus: Check low-scale run in Cirrus CI.

igsilya:
Add a Cirrus CI configuration YAML file with the job to run a minimal
low-scale test.  It is practically not possible or extremely hard to
do the same in GitHub Actions for few reasons:

 - In GHA we're already inside of a container.
 - This container is ubuntu-based.
 - ovn-heater dosn't support debian-based distributions so we need to
   run a nested container.
 - Using a 'container:' config we can run a fedora container, but the
   container command is always overwritten, so no way to run a
   container with systemd as init.
 - I failed to start any container with systemd as init manually.
 - 'runner' is not a root, most of the operations require 'sudo',
   but we will ssh to localhost without using sudo afterwards, even
   if the user is replaced with 'runner' in ansible configuration.
 - Many more environment quirks that doesn't allow you to do what
   you actually need.

Cirrus CI's default containers are similar to GHA, but with Cirrus CI
we're able to just start a plain fedora-cloud VM and have a full access
to it as root.

'runtime' is cached to speed up builds.  Command logging enabled to
simplify troubleshooting.  Container image is cached as well.
'runtime' is archived to avoid symlink breakage.

Since we're re-constructing and re-packing many parts of the 'runtime',
the cache will be re-uploaded every time.  But, at least, we're not
re-building it, which is the main part.

Added a check for already existing venv to allow re-using a cached
version.

Reported-at: https://github.com/dceara/ovn-heater/issues/126

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,57 @@
+low_scale_task:
+
+  compute_engine_instance:
+    image_project: fedora-cloud
+    image: family/fedora-cloud-35
+    platform: linux
+    memory: 8G
+
+  env:
+    DEPENDENCIES: git ansible podman podman-docker
+    PHYS_DEPLOYMENT: ${CIRRUS_WORKING_DIR}/physical-deployments/localhost.yml
+
+  runtime_cache:
+    folder: runtime-cache
+
+  configure_ssh_script:
+    - mkdir -p /root/.ssh/
+    - ssh-keygen -t rsa -N '' -q -f /root/.ssh/id_rsa
+    - ssh-keyscan localhost >> /root/.ssh/known_hosts
+    - cat /root/.ssh/id_rsa.pub >> /root/.ssh/authorized_keys
+    - chmod og-wx /root/.ssh/authorized_keys
+    - ssh root@localhost -v echo Hello
+
+  install_dependencies_script:
+    - dnf install -y ${DEPENDENCIES}
+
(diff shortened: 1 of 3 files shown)


igsilya:
https://cirrus-ci.org/guide/quick-start/

.cirrus.yml
```diff
@@ -0,0 +1,61 @@
...
+    - chmod og-wx /root/.ssh/authorized_keys
+    - ssh root@localhost -v echo Hello
+
+  install_dependencies_script:
+    - dnf install -y ${DEPENDENCIES}
+
+  configure_physical_deployment_script:
+    - rm -rf physical-deployments/physical-deployment.yml
+    - mv physical-deployments/localhost.yml
+         physical-deployments/physical-deployment.yml
```
dceara, line 29: Nit: I'd change this script to:
```
- echo "PHYS_DEPLOYMENT=$(pwd)/physical-deployments/localhost.yml" >> $CIRRUS_ENV
```
in order to point do.sh to the right deployment.

dceara (changes requested):
This is really great!  I just have a minor comment, the rest looks fine to me.

.cirrus.yml
dceara: Yep, sounds good.

dceara (approved):
Thanks!

PR #129, opened 2022-08-31
do.sh: Don't errexit in the post-processing stage.

dceara:
If we managed to run the test and collect its logs it's good enough to
determine if the test passed or failed.  The post-processing stage
shouldn't fail but in case it does (e.g., because of a bug in the
processing tools) it should not trigger the whole test to fail.

Run the post-processing in a separate subshell with errexit disabled.

### DIFF
--- a/do.sh
+++ b/do.sh
@@ -277,36 +277,13 @@ function record_test_config() {
     done
 }
 
-function run_test() {
-    local test_file=$1
-    local out_dir=$2
-    shift; shift
+function mine_data() {
+    out_dir=$1
 
-    # Make sure results dir exists.
-    mkdir -p ${out_dir}/logs
+    echo "-- Mining data from logs in: ${out_dir}"
 
-    # Record SHAs of all components.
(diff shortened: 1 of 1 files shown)


do.sh
```diff
@@ -337,12 +314,51 @@ function run_test() {
...
     # Collecting stats only for 3 workers to avoid bloating the report.
     resource_usage_logs=$(find ${out_dir}/logs -name process-stats.json \
-                            | grep ovn-scale | head -3 || true)
+                            | grep ovn-scale | head -3)
     python3 ${topdir}/utils/process-stats.py \
         resource-usage-report-worker.html ${resource_usage_logs}
+    popd
+}
 
-    deactivate
```
igsilya, line 344: We should, probably, still deactivate somewhere...

dceara: Oops, sorry about that.  Fixed now.

igsilya (approved):
LGTM

PR #131, opened 2022-09-20
test-scenarios: Add 500-node cases.

igsilya:
'cluster-density' scenarios are a bit uneven, e.g. it jumps from 1000 runs to 4000 between 120 and 250 node.  Just doubling that value for 500-node case.

Other tests just scaled linearly according to values in smaller ones.

Signed-off-by: Ilya Maximets \<i.maximets@ovn.org\>


dceara (approved):
Looks good to me, thanks!

PR #134, opened 2022-09-21
density_heavy: service_route: Align VIP subnet size to the one in ovn-tester.

dceara:


dceara:
CC: @igsilya

igsilya (approved):
Looks OK.  Didn't test though, since it takes way too long.

dceara:
> Looks OK. Didn't test though, since it takes way too long.

Thanks!  Let's see how the next scale run looks like.  Hopefully this does it.  Merging for now.

PR #135, opened 2022-09-22
density_heavy: service_route: Iterate inside the VIP subnet when allocating VIPs.

dceara:


dceara:
Pending successful CI.

dceara:
@igsilya this should be ready for review and should now properly fix the issue.  The VIPs are now generated sequentially from within the VIP subnet.

igsilya (approved):
I tested a density-heavy scenario with 120 nodes and it seems to work fine.  LGTM.

PR #136, opened 2022-09-22
Switch from ovn-nbctl, ovn-sbctl, and ovs-vsctl to using ovsdbapp

putnopvut:
As the title says, this series of commits switches the vast majority of calls to *ctl utilities to using ovsdbapp, a python library that wraps the python OVSDB IDL provided by OVS.  It would be apt to compare ovsdbapp to utilities like ovn-nbctl, since it provides shortcut methods for modifying the database.

This marks a change where, instead of having to ssh into the hosts, the ovn-tester code now needs to be able to connect directly to the OVSDB servers running in the containers. The best way to do this is to containerize the ovn-tester and bridge that container to the rest of the containers. In principal, this is incredibly simple, but in order for things to go smoothly, there was a lot of extra stuff that had to be done. Even if this PR is denied, I would argue that containerizing the tester has merits on its own beyond the ability to use an IDL client.

Unfortunately, this change does not eliminate SSH usage entirely. We still use `SSH + docker exec` for a few things:
* On the worker nodes, we have to issue an `ovs-appctl` command in order to make the database listen for incoming connections. If `enable_ssl` is true, then we also have to issue an `ovs-vsctl` command to populate the SSL record.
* We still use SSH in order to set up network namespaces inside the worker containers.
* We still use SSH for `ext_cmd`s configured in test scenarios.

The database transactions are translated 1-to-1 in this PR. In other words, if we performed a single action in an `ovn-nbctl` command, for instance, we now are creating a transaction that does one command. However, since this change introduces first-class Transaction objects, it allows for easily combining multiple commands into a transaction.

Regarding performance, I have done tests of `ovn-low-scale.yml` and `ocp-20-density-heavy.yml` on a single machine. My findings are that ovsdbapp performs similarly to the previous method, especially when using the C JSON parser (https://github.com/ovn-org/ovn-fake-multinode/pull/69 makes it so that OVS built by ovn-fake-multinode will automatically use the C JSON parser) . However, it is possible that certain operations within ovsdbapp may be optimized further.

Picking up new optimizations is one reason that we are pulling from the `master` branch of ovsdbapp when installing it in the tester container. However, the other reason is that at the time of this PR, there is a bug fix that is not in any current releases that allow for more graceful operation when a RAFT leader election takes place.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,6 @@ low_scale_task:
     image: family/fedora-cloud-35
     platform: linux
     memory: 8G
-    use_static_ip: true
 
   env:
     DEPENDENCIES: git ansible podman podman-docker
@@ -15,7 +14,10 @@ low_scale_task:
     folder: runtime-cache
 
   configure_ssh_script:
-    - export IP_ADDR=$(ip route get 8.8.8.8 | head -1 | sed 's/.*src \([0-9\.]*\).*/\1/')
+    - |
+      export IP_ADDR=$(ip route get 8.8.8.8 | \
+      head -1 | \
+      sed 's/.*src \([0-9\.]*\).*/\1/')
     - mkdir -p /root/.ssh/
     - ssh-keygen -t rsa -N '' -q -f /root/.ssh/id_rsa
     - ssh-keyscan ${IP_ADDR} >> /root/.ssh/known_hosts
@@ -32,7 +34,10 @@ low_scale_task:
     - tar -xzf runtime-cache/runtime.tar.gz || true
 
   install_script:
-    - export IP_ADDR=$(ip route get 8.8.8.8 | head -1 | sed 's/.*src \([0-9\.]*\).*/\1/')
+    - |
+      export IP_ADDR=$(ip route get 8.8.8.8 | \
+      head -1 | \
+      sed 's/.*src \([0-9\.]*\).*/\1/')
     - 'sed -i "s/<ip>/${IP_ADDR}/g" ${PHYS_DEPLOYMENT}'
     - ./do.sh install
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,16 +14,12 @@ low_scale_task:
     folder: runtime-cache
 
   configure_ssh_script:
-    - |
-      export IP_ADDR=$(ip route get 8.8.8.8 | \
-      head -1 | \
-      sed 's/.*src \([0-9\.]*\).*/\1/')
     - mkdir -p /root/.ssh/
     - ssh-keygen -t rsa -N '' -q -f /root/.ssh/id_rsa
-    - ssh-keyscan ${IP_ADDR} >> /root/.ssh/known_hosts
+    - ssh-keyscan $(hostname) >> /root/.ssh/known_hosts
     - cat /root/.ssh/id_rsa.pub >> /root/.ssh/authorized_keys
     - chmod og-wx /root/.ssh/authorized_keys
-    - ssh root@${IP_ADDR} -v echo Hello
+    - ssh root@$(hostname) -v echo Hello
 
   install_dependencies_script:
     - dnf install -y ${DEPENDENCIES}
@@ -34,11 +30,7 @@ low_scale_task:
     - tar -xzf runtime-cache/runtime.tar.gz || true
 
   install_script:
-    - |
-      export IP_ADDR=$(ip route get 8.8.8.8 | \
-      head -1 | \
-      sed 's/.*src \([0-9\.]*\).*/\1/')
-    - 'sed -i "s/<ip>/${IP_ADDR}/g" ${PHYS_DEPLOYMENT}'
+    - 'sed -i "s/<host>/$(hostname)/g" ${PHYS_DEPLOYMENT}'
     - ./do.sh install
 
   pack_caches_script:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,21 +5,23 @@ low_scale_task:
     image: family/fedora-cloud-35
     platform: linux
     memory: 8G
+    use_static_ip: true
 
   env:
     DEPENDENCIES: git ansible podman podman-docker
-    PHYS_DEPLOYMENT: ${CIRRUS_WORKING_DIR}/physical-deployments/localhost.yml
+    PHYS_DEPLOYMENT: ${CIRRUS_WORKING_DIR}/physical-deployments/ci.yml
 
   runtime_cache:
     folder: runtime-cache
 
   configure_ssh_script:
+    - export IP_ADDR=$(ip route get 8.8.8.8 | head -1 | sed 's/.*src \([0-9\.]*\).*/\1/')
     - mkdir -p /root/.ssh/
     - ssh-keygen -t rsa -N '' -q -f /root/.ssh/id_rsa
-    - ssh-keyscan localhost >> /root/.ssh/known_hosts
+    - ssh-keyscan ${IP_ADDR} >> /root/.ssh/known_hosts
     - cat /root/.ssh/id_rsa.pub >> /root/.ssh/authorized_keys
     - chmod og-wx /root/.ssh/authorized_keys
-    - ssh root@localhost -v echo Hello
+    - ssh root@${IP_ADDR} -v echo Hello
 
   install_dependencies_script:
     - dnf install -y ${DEPENDENCIES}
@@ -30,6 +32,8 @@ low_scale_task:
     - tar -xzf runtime-cache/runtime.tar.gz || true
 
   install_script:
+    - export IP_ADDR=$(ip route get 8.8.8.8 | head -1 | sed 's/.*src \([0-9\.]*\).*/\1/')
+    - 'sed -i "s/<ip>/${IP_ADDR}/g" ${PHYS_DEPLOYMENT}'
     - ./do.sh install
 
   pack_caches_script:
--- a/do.sh
+++ b/do.sh
@@ -304,10 +304,16 @@ function record_test_config() {
 
 function mine_data() {
     out_dir=$1
+    tester_host=$2
 
     echo "-- Mining data from logs in: ${out_dir}"
 
     pushd ${out_dir}
+    # Prior to containerization of ovn-tester, HTML files written by ovn-tester
+    # were written directly to ${out_dir}. To make things easier for tools, we
+    # copy the HTML files back to this original location.
+    cp logs/${tester_host}/ovn-tester/*.html ${out_dir}
+
     mkdir -p mined-data
     for p in ovn-northd ovn-controller ovn-nbctl; do
         logs=$(find ${out_dir}/logs -name ${p}.log)
@@ -403,10 +409,6 @@ function run_test() {
     for f in *.tgz; do
        tar xvfz $f
     done
-    # Prior to containerization of ovn-tester, HTML files written by ovn-tester
-    # were written directly to ${out_dir}. To make things easier for tools, we
-    # copy the HTML files back to this original location.
-    cp ${tester_host}/ovn-tester/*.html ${out_dir}
 
     # Once we successfully ran the test and collected its logs, the post
     # processing (e.g., data mining) can run in a subshell with errexit
@@ -414,7 +416,7 @@ function run_test() {
     # processing fails.
     (
         set +o errexit
-        mine_data ${out_dir}
+        mine_data ${out_dir} ${tester_host}
     )
 }
 
--- a/do.sh
+++ b/do.sh
@@ -309,10 +309,6 @@ function mine_data() {
     echo "-- Mining data from logs in: ${out_dir}"
 
     pushd ${out_dir}
-    # Prior to containerization of ovn-tester, HTML files written by ovn-tester
-    # were written directly to ${out_dir}. To make things easier for tools, we
-    # copy the HTML files back to this original location.
-    cp logs/${tester_host}/ovn-tester/*.html ${out_dir}
 
     mkdir -p mined-data
     for p in ovn-northd ovn-controller ovn-nbctl; do
@@ -409,6 +405,10 @@ function run_test() {
     for f in *.tgz; do
        tar xvfz $f
     done
+    # Prior to containerization of ovn-tester, HTML files written by ovn-tester
+    # were written directly to ${out_dir}. To make things easier for tools, we
+    # copy the HTML files back to this original location.
+    cp logs/${tester_host}/ovn-tester/*.html ${out_dir} || true
 
     # Once we successfully ran the test and collected its logs, the post
     # processing (e.g., data mining) can run in a subshell with errexit
--- a/do.sh
+++ b/do.sh
@@ -17,6 +17,8 @@ ovn_fmn_playbooks=${ovn_fmn_utils}/playbooks
 ovn_fmn_generate=${ovn_fmn_utils}/generate-hosts.py
 ovn_fmn_docker=${ovn_fmn_utils}/generate-docker-cfg.py
 ovn_fmn_podman=${ovn_fmn_utils}/generate-podman-cfg.py
+ovn_fmn_get=${ovn_fmn_utils}/get-config-value.py
+ovn_fmn_ip=${rundir}/ovn-fake-multinode/ip_gen.py
 hosts_file=${rundir}/hosts
 installer_log_file=${rundir}/installer-log
 docker_daemon_file=${rundir}/docker-daemon.json
@@ -26,7 +28,6 @@ log_perf_file=${rundir}/perf.sh
 process_monitor_file=${rundir}/process-monitor.py
 
 ovn_tester=${topdir}/ovn-tester
-ovn_tester_log_file=test-log
 
 EXTRA_OPTIMIZE=${EXTRA_OPTIMIZE:-no}
 USE_OVSDB_ETCD=${USE_OVSDB_ETCD:-no}
@@ -92,7 +93,7 @@ function install_venv() {
         python3 -m virtualenv ${ovn_heater_venv}
     fi
     source ${ovn_heater_venv}/bin/activate
-    pip install -r ${ovn_tester}/requirements.txt
+    pip install -r ${topdir}/utils/requirements.txt
     deactivate
     popd
 }
@@ -228,6 +229,22 @@ function install_ovn_fake_multinode() {
     popd
 }
 
+function install_ovn_tester() {
+    rm -rf tester_files
+    mkdir tester_files
+    ssh_key=$(${ovn_fmn_get} ${phys_deployment} tester-node ssh_key)
+    # We need to copy the files into a known directory within the Docker
+    # context directory. Otherwise, Docker can't find the files we reference.
+    cp ${ssh_key} tester_files
+    cp ${phys_deployment} tester_files
+    ssh_key_file=tester_files/$(basename ${ssh_key})
+    phys_deployment_file=tester_files/$(basename ${phys_deployment})
+    docker build -t ovn/ovn-tester --build-arg SSH_KEY=${ssh_key_file} --build-arg PHYS_DEPLOYMENT=${phys_deployment_file} -f ${topdir}/Dockerfile .
+    docker tag ovn/ovn-tester localhost:5000/ovn/ovn-tester
+    docker push localhost:5000/ovn/ovn-tester
+    rm -rf tester_files
+}
+
 # Prepare OVS bridges and cleanup containers.
 function init_ovn_fake_multinode() {
     echo "-- Initializing ovn-fake-multinode cluster on all nodes"
@@ -240,6 +257,10 @@ function pull_ovn_fake_multinode() {
     ansible-playbook ${ovn_fmn_playbooks}/pull-fake-multinode.yml -i ${hosts_file}
 }
 
+function pull_ovn_tester() {
+    ansible-playbook ${ovn_fmn_playbooks}/pull-ovn-tester.yml -i ${hosts_file}
+}
+
 function install() {
     pushd ${rundir}
     install_deps
@@ -248,6 +269,10 @@ function install() {
     install_ovn_fake_multinode
     init_ovn_fake_multinode
     pull_ovn_fake_multinode
+    popd
+    pushd ${topdir}
+    install_ovn_tester
+    pull_ovn_tester
     popd
 }
 
@@ -282,7 +307,6 @@ function mine_data() {
 
     echo "-- Mining data from logs in: ${out_dir}"
 
-
     pushd ${out_dir}
     mkdir -p mined-data
     for p in ovn-northd ovn-controller ovn-nbctl; do
@@ -301,6 +325,7 @@ function mine_data() {
     grep ovn-installed ${logs} | cut -d ':' -f 2- | tr '|' ' ' \
         | cut -d ' ' -f 1,7 | tr 'T' ' ' | sort > mined-data/ovn-installed.log
 
+    source ${rundir}/${ovn_heater_venv}/bin/activate
     python3 ${topdir}/utils/latency.py "$(date +%z)" \
         ./mined-data/ovn-binding.log ./mined-data/ovn-installed.log \
         > mined-data/binding-to-ovn-installed-latency
@@ -321,7 +346,30 @@ function mine_data() {
                             | grep ovn-scale | head -3)
     python3 ${topdir}/utils/process-stats.py \
         resource-usage-report-worker.html ${resource_usage_logs}
+    deactivate
+
     popd
+}
+
+function get_tester_ip() {
+    local test_file=$1
+
+    node_net=$(${ovn_fmn_get} ${test_file} cluster node_net --default=192.16.0.0/16)
+    node_cidr=${node_net#*/}
+    node_ip=${node_net%/*}
+    ip_index=1
+    clustered_db=$(${ovn_fmn_get} ${test_file} cluster clustered_db --default=True)
+    # The yaml loader normalizes "yes" and "no" values for clustered_db into
+    # "True" and "False"
+    if [ "${clustered_db}" = "True" ]; then
+        (( ip_index += 3 ))
+    else
+        (( ip_index += 1 ))
+    fi
+    n_relays=$(${ovn_fmn_get} ${test_file} cluster n_relays --default=0)
+    (( ip_index += ${n_relays} ))
+    tester_ip=$(${ovn_fmn_ip} ${node_net} ${node_ip} ${ip_index})
+    echo "${tester_ip}/${node_cidr}"
 }
 
 function run_test() {
@@ -338,11 +386,14 @@ function run_test() {
     # Perform a fast cleanup by doing a minimal redeploy.
     init_ovn_fake_multinode
 
-    source ${rundir}/${ovn_heater_venv}/bin/activate
-    pushd ${out_dir}
+    tester_ip=$(get_tester_ip ${test_file})
+    if ! ansible-playbook ${ovn_fmn_playbooks}/run-tester.yml -i ${hosts_file} --extra-vars "test_file=${test_file} tester_ip=${tester_ip}" ; then
+         echo "-- Failed to set up test!"
+    fi
 
-    if ! python -u ${ovn_tester}/ovn_tester.py $phys_deployment ${test_file} 2>&1 | tee ${ovn_tester_log_file}; then
-        echo "-- Failed to run test! Check logs at: $PWD/${ovn_tester_log_file}"
+    tester_host=$(${ovn_fmn_get} ${phys_deployment} tester-node name)
+    if ! ssh root@${tester_host} docker exec ovn-tester python3 -u /ovn-tester/ovn_tester.py /physical-deployment.yml /test-scenario.yml 2>&1 | tee ${out_dir}/test-log ; then
+        echo "-- Failed to run test. Check logs at: ${out_dir}/test-log"
     fi
 
     echo "-- Collecting logs to: ${out_dir}"
@@ -352,7 +403,10 @@ function run_test() {
     for f in *.tgz; do
        tar xvfz $f
     done
-    popd
+    # Prior to containerization of ovn-tester, HTML files written by ovn-tester
+    # were written directly to ${out_dir}. To make things easier for tools, we
+    # copy the HTML files back to this original location.
+    cp ${tester_host}/ovn-tester/*.html ${out_dir}
 
     # Once we successfully ran the test and collected its logs, the post
     # processing (e.g., data mining) can run in a subshell with errexit
@@ -362,9 +416,6 @@ function run_test() {
         set +o errexit
         mine_data ${out_dir}
     )
-
-    popd
-    deactivate
 }
 
 function usage() {
--- a/do.sh
+++ b/do.sh
@@ -356,20 +356,11 @@ function mine_data() {
 function get_tester_ip() {
     local test_file=$1
 
+    # The tester gets the first IP address in the configured node_net.
     node_net=$(${ovn_fmn_get} ${test_file} cluster node_net --default=192.16.0.0/16)
     node_cidr=${node_net#*/}
     node_ip=${node_net%/*}
     ip_index=1
-    clustered_db=$(${ovn_fmn_get} ${test_file} cluster clustered_db --default=True)
-    # The yaml loader normalizes "yes" and "no" values for clustered_db into
-    # "True" and "False"
-    if [ "${clustered_db}" = "True" ]; then
-        (( ip_index += 3 ))
-    else
-        (( ip_index += 1 ))
-    fi
-    n_relays=$(${ovn_fmn_get} ${test_file} cluster n_relays --default=0)
-    (( ip_index += ${n_relays} ))
     tester_ip=$(${ovn_fmn_ip} ${node_net} ${node_ip} ${ip_index})
     echo "${tester_ip}/${node_cidr}"
 }
--- a/do.sh
+++ b/do.sh
@@ -3,7 +3,8 @@
 set -o errexit -o pipefail
 
 topdir=$(pwd)
-rundir=${topdir}/runtime
+rundir_name=runtime
+rundir=${topdir}/${rundir_name}
 deployment_dir=${topdir}/physical-deployments
 result_dir=${topdir}/test_results
 
@@ -230,17 +231,14 @@ function install_ovn_fake_multinode() {
 }
 
 function install_ovn_tester() {
-    rm -rf tester_files
-    mkdir tester_files
     ssh_key=$(${ovn_fmn_get} ${phys_deployment} tester-node ssh_key)
     # We need to copy the files into a known directory within the Docker
     # context directory. Otherwise, Docker can't find the files we reference.
-    cp ${ssh_key} tester_files
-    ssh_key_file=tester_files/$(basename ${ssh_key})
-    docker build -t ovn/ovn-tester --build-arg SSH_KEY=${ssh_key_file} -f ${topdir}/Dockerfile .
+    cp ${ssh_key} .
+    ssh_key_file=${rundir_name}/$(basename ${ssh_key})
+    docker build -t ovn/ovn-tester --build-arg SSH_KEY=${ssh_key_file} -f ${topdir}/Dockerfile ${topdir}
     docker tag ovn/ovn-tester localhost:5000/ovn/ovn-tester
     docker push localhost:5000/ovn/ovn-tester
-    rm -rf tester_files
 }
 
 # Prepare OVS bridges and cleanup containers.
@@ -267,8 +265,6 @@ function install() {
     install_ovn_fake_multinode
     init_ovn_fake_multinode
     pull_ovn_fake_multinode
-    popd
-    pushd ${topdir}
     install_ovn_tester
     pull_ovn_tester
     popd
--- a/do.sh
+++ b/do.sh
@@ -332,7 +332,7 @@ function mine_data() {
         | cut -d ' ' -f 1,7 | tr 'T' ' ' | sort > mined-data/ovn-installed.log
 
     source ${rundir}/${ovn_heater_venv}/bin/activate
-    python3 ${topdir}/utils/latency.py "$(date +%z)" \
+    python3 ${topdir}/utils/latency.py \
         ./mined-data/ovn-binding.log ./mined-data/ovn-installed.log \
         > mined-data/binding-to-ovn-installed-latency
 
--- a/do.sh
+++ b/do.sh
@@ -397,7 +397,7 @@ function run_test() {
     # Prior to containerization of ovn-tester, HTML files written by ovn-tester
     # were written directly to ${out_dir}. To make things easier for tools, we
     # copy the HTML files back to this original location.
-    cp logs/${tester_host}/ovn-tester/*.html ${out_dir} || true
+    cp ${tester_host}/ovn-tester/*.html ${out_dir} || true
 
     # Once we successfully ran the test and collected its logs, the post
     # processing (e.g., data mining) can run in a subshell with errexit
--- a/do.sh
+++ b/do.sh
@@ -236,10 +236,8 @@ function install_ovn_tester() {
     # We need to copy the files into a known directory within the Docker
     # context directory. Otherwise, Docker can't find the files we reference.
     cp ${ssh_key} tester_files
-    cp ${phys_deployment} tester_files
     ssh_key_file=tester_files/$(basename ${ssh_key})
-    phys_deployment_file=tester_files/$(basename ${phys_deployment})
-    docker build -t ovn/ovn-tester --build-arg SSH_KEY=${ssh_key_file} --build-arg PHYS_DEPLOYMENT=${phys_deployment_file} -f ${topdir}/Dockerfile .
+    docker build -t ovn/ovn-tester --build-arg SSH_KEY=${ssh_key_file} -f ${topdir}/Dockerfile .
     docker tag ovn/ovn-tester localhost:5000/ovn/ovn-tester
     docker push localhost:5000/ovn/ovn-tester
     rm -rf tester_files
@@ -380,7 +378,7 @@ function run_test() {
     init_ovn_fake_multinode
 
     tester_ip=$(get_tester_ip ${test_file})
-    if ! ansible-playbook ${ovn_fmn_playbooks}/run-tester.yml -i ${hosts_file} --extra-vars "test_file=${test_file} tester_ip=${tester_ip}" ; then
+    if ! ansible-playbook ${ovn_fmn_playbooks}/run-tester.yml -i ${hosts_file} --extra-vars "test_file=${test_file} tester_ip=${tester_ip} phys_deployment=${phys_deployment}" ; then
          echo "-- Failed to set up test!"
     fi
 
--- a/do.sh
+++ b/do.sh
@@ -375,7 +375,7 @@ function run_test() {
 
     tester_ip=$(get_tester_ip ${test_file})
     if ! ansible-playbook ${ovn_fmn_playbooks}/run-tester.yml -i ${hosts_file} --extra-vars "test_file=${test_file} tester_ip=${tester_ip} phys_deployment=${phys_deployment}" ; then
-         echo "-- Failed to set up test!"
+        die "-- Failed to set up test!"
     fi
 
     tester_host=$(${ovn_fmn_get} ${phys_deployment} tester-node name)
--- a/ovn-fake-multinode-utils/get-config-value.py
+++ b/ovn-fake-multinode-utils/get-config-value.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 import argparse
 import yaml
 
--- a/ovn-fake-multinode-utils/get-config-value.py
+++ b/ovn-fake-multinode-utils/get-config-value.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+import argparse
+import yaml
+
+
+def parser_setup(parser):
+    group = parser.add_argument_group()
+    group.add_argument(
+        "config",
+        metavar="CONFIG_FILE",
+        help="Configuration file",
+    )
+    group.add_argument(
+        "section",
+        metavar="SECTION",
+        help="Configuration file section",
+    )
+    group.add_argument(
+        "variable", metavar="VAR", help="Configuration variable to retrieve"
+    )
+    group.add_argument(
+        "--default", help="Default value if VAR is not in CONFIG"
+    )
+
+
+def get_config_value(args):
+    with open(args.config, 'r') as config_file:
+        parsed = yaml.safe_load(config_file)
+
+    try:
+        return parsed[args.section][args.variable]
+    except KeyError:
+        if args.default:
+            return args.default
+        raise
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Read YAML config value")
+    parser_setup(parser)
+    args = parser.parse_args()
+    val = get_config_value(args)
+    print(val, end='')
+
+
+if __name__ == "__main__":
+    main()
--- a/ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
+++ b/ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
@@ -11,6 +11,15 @@
     shell: |
       docker volume prune -f
 
+  - name: Stop tester container if running
+    when: ovn_tester is defined
+    ignore_errors: yes
+    shell: |
+      cd {{ ovn_fake_multinode_target_path }}/ovn-fake-multinode
+      ./ovs-docker del-ports br-ovn ovn-tester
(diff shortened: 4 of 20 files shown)


.cirrus.yml
```diff
@@ -30,6 +31,8 @@ low_scale_task:
     - tar -xzf runtime-cache/runtime.tar.gz || true
 
   install_script:
+    - export IP_ADDR=$(ip route get 8.8.8.8 | head -1 | sed 's/.*src \([0-9\.]*\).*/\1/')
+    - 'sed -i "s/<ip>/${IP_ADDR}/g" ${PHYS_DEPLOYMENT}'
```
igsilya, line 35: You also need to change the `configure_ssh_script` step, because it only configures SSH for `localhost`.

igsilya:
That looks very promising.  Thanks, @putnopvut !

I didn't look close at the code, but I have one high-level comment after looking at the CI logs.
I think, `log_cmds` should enable jsonrpc debug logs or some other type of verbose logging in ovsdbapp, otherwise it will be hard to debug test scenarios.  What do you think?

Another high level comment is that with this change we no longer have a live stream of the test progress.  That is unfortunate, but it looks like ansible doesn't support streaming of the command output, even though this feature is long requested (https://github.com/ansible/proposals/issues/92).  It's not a huge problem, I guess.

putnopvut:
> That looks very promising. Thanks, @putnopvut !
> 
> I didn't look close at the code, but I have one high-level comment after looking at the CI logs. I think, `log_cmds` should enable jsonrpc debug logs or some other type of verbose logging in ovsdbapp, otherwise it will be hard to debug test scenarios. What do you think?

It sounds good. I'll have to look into this to find out what this would entail. If I were guessing, I'm guessing you get verbose logging from ovsdbapp by changing the level of the logger to DEBUG. The only downside of this is that it will also give you low-level output from paramiko and other libraries. I'll see what I can work out.
 
> Another high level comment is that with this change we no longer have a live stream of the test progress. That is unfortunate, but it looks like ansible doesn't support streaming of the command output, even though this feature is long requested ([ansible/proposals#92](https://github.com/ansible/proposals/issues/92)). It's not a huge problem, I guess.

Yes, I mentioned this in commit #3 in this series. I was disappointed to find that ansible cannot stream output in realtime. It appears this has been proposed several times in the ansible project and it's a difficult problem to fix, mainly because it becomes the responsibility of all ansible submodules to support this, rather than a single global top-level change.

igsilya:
> I'm guessing you get verbose logging from ovsdbapp by changing the level of the logger to DEBUG. The only downside of this is that it will also give you low-level output from paramiko and other libraries. I'll see what I can work out.

It looks like ovsdbapp is using named loggers, so something like this should, probably, work:
```
logging.getLogger("transaction").setLevel(logging.DEBUG)
```

> I was disappointed to find that ansible cannot stream output in realtime. It appears this has been proposed several times in the ansible project and it's a difficult problem to fix, mainly because it becomes the responsibility of all ansible submodules to support this, rather than a single global top-level change.

Looking at the ovn-tester playbook, I'm not sure what is an actual value of running the test itself from ansible.  It only provides an ssh connection, but we still need to call docker manually.  Maybe we can keep the configuration/startup in ansible, but call the actual test directly?  E.g.:
```
ssh root@${tester_node} docker exec ovn-tester \
    python3 -u /ovn-tester/ovn_tester.py /physical-deployment.yml /test-scenario.yml
```
This way we can still see the log in real time and even store it locally with tee as we do now.

igsilya (commented):
> Despite the fact that the containerized tester is intended to
> communicate directly with the DBs on the test nodes, there is still a
> need for the tester to ssh to the hosts. This is because we still need
> to be able to do non-DB operations, such as creating network namespaces.

This is not entirely true.  Tester doesn't need access to physical hosts in order to create namespaces inside containers.  It just needs a direct SSH connection to all containers to do that.

> As such, the tester needs to be given an SSH private key that can be
> used to communicate with the hosts. This is configured in the physical
> deployment configuration. Perhaps a future improvement could eliminate
> all SSH usage in ovn-tester, but for now it's still necessary.

Perhaps, we just need to go "all in" with the containerization and avoid impersonating the orchestrator.

I spent some time with ansible and came up with the following playbook: https://github.com/igsilya/ovn-heater/commit/816f67048d074a3be30c5dc5ad113166031bb395 .  This playbook is able to bring up all the containers we need and wire them up for passwordless SSH from the tester container.

Having that, we should be able to remove the usage of `./ovn_cluster.sh` and the process monitor script from the ovn-tester code.  For all the other operations we should not need an access to physical hosts, IIUC.  ovn-tester can be converted to establish direct SSH connections to all containers and not mess with a docker and ugly 'invoke_shell' calls.

What do you think?

It's quiet a bit of work cleaning up ovn-tester, and we'll also need a good solution to not duplicate defaults, e.g. get the test config in the orchestrator, apply all the defaults for all the variables that are not specified and pass a full yml with no omitted values to the tester, so we will not decide what the default value for the mgmt_net, for example, in two different places.
Saying that, it might make sense to do that as a separate PR or we can integrate that work into current PR to not touch the same code twice.

putnopvut:
> > I'm guessing you get verbose logging from ovsdbapp by changing the level of the logger to DEBUG. The only downside of this is that it will also give you low-level output from paramiko and other libraries. I'll see what I can work out.
> 
> It looks like ovsdbapp is using named loggers, so something like this should, probably, work:
> 
> ```
> logging.getLogger("transaction").setLevel(logging.DEBUG)
> ```
> 

OK, this seems easy enough to at least follow up on. I can test it out and see what sort of output we end up getting.

> > I was disappointed to find that ansible cannot stream output in realtime. It appears this has been proposed several times in the ansible project and it's a difficult problem to fix, mainly because it becomes the responsibility of all ansible submodules to support this, rather than a single global top-level change.
> 
> Looking at the ovn-tester playbook, I'm not sure what is an actual value of running the test itself from ansible. It only provides an ssh connection, but we still need to call docker manually. Maybe we can keep the configuration/startup in ansible, but call the actual test directly? E.g.:
> 
> ```
> ssh root@${tester_node} docker exec ovn-tester \
>     python3 -u /ovn-tester/ovn_tester.py /physical-deployment.yml /test-scenario.yml
> ```
> 
> This way we can still see the log in real time and even store it locally with tee as we do now.

This sounds reasonable to me. I'll give it a try and see how it works out.

putnopvut:
> > Despite the fact that the containerized tester is intended to
> > communicate directly with the DBs on the test nodes, there is still a
> > need for the tester to ssh to the hosts. This is because we still need
> > to be able to do non-DB operations, such as creating network namespaces.
> 
> This is not entirely true. Tester doesn't need access to physical hosts in order to create namespaces inside containers. It just needs a direct SSH connection to all containers to do that.
> 
> > As such, the tester needs to be given an SSH private key that can be
> > used to communicate with the hosts. This is configured in the physical
> > deployment configuration. Perhaps a future improvement could eliminate
> > all SSH usage in ovn-tester, but for now it's still necessary.
> 
> Perhaps, we just need to go "all in" with the containerization and avoid impersonating the orchestrator.
> 
> I spent some time with ansible and came up with the following playbook: [igsilya@816f670](https://github.com/igsilya/ovn-heater/commit/816f67048d074a3be30c5dc5ad113166031bb395) . This playbook is able to bring up all the containers we need and wire them up for passwordless SSH from the tester container.
> 
> Having that, we should be able to remove the usage of `./ovn_cluster.sh` and the process monitor script from the ovn-tester code. For all the other operations we should not need an access to physical hosts, IIUC. ovn-tester can be converted to establish direct SSH connections to all containers and not mess with a docker and ugly 'invoke_shell' calls.
> 
> What do you think?
> 
> It's quiet a bit of work cleaning up ovn-tester, and we'll also need a good solution to not duplicate defaults, e.g. get the test config in the orchestrator, apply all the defaults for all the variables that are not specified and pass a full yml with no omitted values to the tester, so we will not decide what the default value for the mgmt_net, for example, in two different places. Saying that, it might make sense to do that as a separate PR or we can integrate that work into current PR to not touch the same code twice.

I like the idea of starting all containers before executing the test. That way, IP address management for the containers can all be orchestrated prior to running the test. It would make for some less hack-y IP address calculations in general, I think. I also really like the idea of setting up SSH connectivity in the playbook before running the test. It allows for you to generate the SSH keys as part of the playbook, so there is no need to create SSH keys beforehand, and there is no need to tell the physical deployment where this SSH key is.

The bigger question is whether SSHing directly to containers is actually a net positive. The number of SSH connections from the tester container would jump from a relatively low number (40 at most?) to the number of nodes under test. So for a 120 node test, the tester container would now potentially have 120 SSH connections open. I actually have similar concerns about the connections from the tester to the OVS database running on each container.

igsilya:
> The bigger question is whether SSHing directly to containers is actually a net positive. The number of SSH connections from the tester container would jump from a relatively low number (40 at most?) to the number of nodes under test. So for a 120 node test, the tester container would now potentially have 120 SSH connections open. I actually have similar concerns about the connections from the tester to the OVS database running on each container.

I think, it should not be a problem, unless we ran out of open file descriptors, but that should not be a big problem either, we can always change the limit in the tester container.  Our databases have sustained SSL/TLS connections to all ovn-controllers, which is the same number, and they handle that just fine.
I guess, we need to test and find out, but architecturally it looks like a better solution to clearly separate the tester from the infrastructure.

putnopvut:
> > It looks like ovsdbapp is using named loggers, so something like this should, probably, work:
> > ```
> > logging.getLogger("transaction").setLevel(logging.DEBUG)
> > ```
> 
> OK, this seems easy enough to at least follow up on. I can test it out and see what sort of output we end up getting.
>

It took some work, but I got this working finally. The problem is that the logging name is not just "transaction" as I would expect, but rather "ovsdbapp.backend.ovs_idl.transaction". I will include this in the next version of the PR. I'll wait until I have everything else implemented first.

putnopvut:
> > Looking at the ovn-tester playbook, I'm not sure what is an actual value of running the test itself from ansible. It only provides an ssh connection, but we still need to call docker manually. Maybe we can keep the configuration/startup in ansible, but call the actual test directly? E.g.:
> > ```
> > ssh root@${tester_node} docker exec ovn-tester \
> >     python3 -u /ovn-tester/ovn_tester.py /physical-deployment.yml /test-scenario.yml
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This way we can still see the log in real time and even store it locally with tee as we do now.
> 
> This sounds reasonable to me. I'll give it a try and see how it works out.

This worked as expected. It actually made things ever so slightly faster since I didn't have to compress, copy, and then uncompress the ovn-tester log file.

putnopvut:
I just pushed a new version of the series. The differences are:
* We don't run the test in ansible now. Instead, we just use ssh+python to do it. This allows for realtime logging to the console during a test, and it allows for us to `tee` the output to the `test-log` file directly. It also has a slight performance benefit since we do not have to tar, copy, and then untar the tester log file.
* The global `log_cmds` option now logs ovsdbapp transactions, similar to what you would see when `ovn-nbctl` was used.

What I have not addressed is the idea of replacing SSHing to the hosts with SSHing to the containers directly. IMO, this would probably work better as a follow-up PR.

putnopvut:
I'm aware of the failing check. The issue is that the `test-log` file is not being written. There's something wrong with:
````
    if ! ssh root@${tester_host} docker exec "ovn-tester python3 -u /ovn-tester/ovn_tester.py /physical-deployment.yml /test-scenario.yml" 2>&1 | tee ${ovn_tester_log_file} ; then
        echo "-- Failed to run test. Check logs at: ${ovn_tester_log_file}"
    fi
````
I thought the issue was that `| tee ...` was being interpreted as part of the `docker exec` command, and so the write was happening inside the container. However, `test-log` cannot be found in the `ovn-tester` container after a test run, and I don't see any error output from `tee`. Once I get this figured out, I'll push changes that fix the failing check.

File: do.sh [hunk 2]
```diff
@@ -228,6 +229,22 @@ function install_ovn_fake_multinode() {
     popd
 }
 
+function install_ovn_tester() {
+    rm -rf tester_files
+    mkdir tester_files
+    ssh_key=$(${ovn_fmn_get} ${phys_deployment} tester-node ssh_key)
```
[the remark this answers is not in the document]
dceara: After some more trial and error, looks like `ssh_key` is the right key but instead we should change the physical-deployment/physical-deployment.yml file.

do.sh
```diff
@@ -228,6 +229,22 @@ function install_ovn_fake_multinode() {
...
+    # We need to copy the files into a known directory within the Docker
+    # context directory. Otherwise, Docker can't find the files we reference.
+    cp ${ssh_key} tester_files
+    cp ${phys_deployment} tester_files
+    ssh_key_file=tester_files/$(basename ${ssh_key})
+    phys_deployment_file=tester_files/$(basename ${phys_deployment})
+    docker build -t ovn/ovn-tester --build-arg SSH_KEY=${ssh_key_file} --build-arg PHYS_DEPLOYMENT=${phys_deployment_file} -f ${topdir}/Dockerfile .
```
dceara, line 242: I know we mention in the readme that ovn-heater should be used with care.  But by default we copy the current user's private SSH key to a container image that will likely run on a different machine.  It makes me feel a bit uneasy.

ovn-fake-multinode-utils/playbooks/run-tester.yml
```diff
@@ -0,0 +1,20 @@
...
+  tasks:
+  - name: Start tester container
+    when: ovn_tester is defined
+    shell: |
+      docker run -dt --name=ovn-tester --hostname=ovn-tester --privileged ovn/ovn-tester
+
+  - name: Copy files to the tester container
+    when: ovn_tester is defined
+    shell: |
+      docker cp {{ test_file }} ovn-tester:/test-scenario.yml
```
dceara, line 13: This only works if the tester is the same machine as the orchestrator.  Otherwise we need to move the test file to the tester too.

dceara (commented):
Thanks for the PR @putnopvut this is great!  I didn't do a full review yet so I just left some notes from when I was trying it out on our scale setup.

do.sh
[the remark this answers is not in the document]
putnopvut: Yes, sorry for the confusion. It is intended to be `ssh_key` and not `ssh-key`. Sounds like the default YAML file needs updating like you said.

do.sh
putnopvut: OK, it sounds like the best course of action for now would be not to have a default SSH key location. Users of ovn-heater will need to generate one (or point to an existing one) manually.

BTW, if we take up Ilya's suggestion of pre-creating the containers and then SSHing directly to those containers during the test, then all of this will be moot.

ovn-fake-multinode-utils/playbooks/run-tester.yml
putnopvut: Ah, good catch.

do.sh
```diff
@@ -249,6 +270,10 @@ function install() {
     init_ovn_fake_multinode
     pull_ovn_fake_multinode
     popd
+    pushd ${topdir}
+    install_ovn_tester
+    pull_ovn_tester
```
dceara, line 275: Does it make sense to move `install_ovn_tester` and `pull_ovn_tester` to `run_test()`?  That way we don't have to run `./do.sh install` every time the tester code changed.

dceara: Actually I was wrong:
```
STEP 8/8: RUN pip3 install -r /ovn-tester/requirements.txt
--> Using cache 9e62908ad117c9d2ee2f3192ebc38f317064b931a38c4d11ac8203f5f626e602
```
So i'm not sure what's best. :)

putnopvut: When I first started this process, I had `install_ovn_tester` and `pull_ovn_tester` as part of `run_test()`. I ended up changing it because my workflow would go like so:
- Make adjustments to `do.sh`
- Run a test
- Wait forever for the tester container to be built
- See a test failure.
- Go to 1

That third step was causing a bunch of stress, and I realized that a lot of the tester container is constant and known at install-time, so I adjusted to building the docker image at install time and running it at runtime.

As you point out, now if there is an error in the tester code, I have to reinstall to re-test it. But if there is a problem in a config file or `do.sh`, re-running the test is fast.

do.sh
```diff
@@ -301,12 +325,12 @@ function mine_data() {
     grep ovn-installed ${logs} | cut -d ':' -f 2- | tr '|' ' ' \
         | cut -d ' ' -f 1,7 | tr 'T' ' ' | sort > mined-data/ovn-installed.log
 
+    source ${rundir}/${ovn_heater_venv}/bin/activate
     python3 ${topdir}/utils/latency.py "$(date +%z)" \
```
igsilya, line 329: The `date +%z` here takes the time zone of the orchestrator, but the value will be used to offset the time taken from the log on the tester node.  This results in very large negative latency values in the mined data.  We should either:

- Get the time zone from the tester container
- Synchronize time zones between the orchestrator and the tester container
- Make ovn-tester to use UTC in logs and remove the time adjustments from the latency calculations.

do.sh
```diff
@@ -348,7 +398,10 @@ function run_test() {
     for f in *.tgz; do
        tar xvfz $f
     done
-    popd
+    # Prior to containerization of ovn-tester, HTML files written by ovn-tester
+    # were written directly to ${out_dir}. To make things easier for tools, we
+    # copy the HTML files back to this original location.
+    cp ${tester_host}/ovn-tester/*.html ${out_dir}
```
igsilya, line 404: If the test failed and there are no html files, this command will fail the whole ./do.sh because we're running with errexit.  This way we will loose the data mining stage.

ovn-tester/ovn_utils.py
```diff
@@ -38,6 +46,9 @@
 DualStackIP = namedtuple('DualStackIP', ['ip4', 'plen4', 'ip6', 'plen6'])
 
 
+vlog.use_python_logger(max_level=vlog.INFO)
```
igsilya, line 49: There is one more thing we need to do.  Python-IDL is using select.select to simulate select.poll by default to avoid broken multithreading with select.poll implementation vs eventlet/gevent.  Unfortunately, that doesn't work if we're trying to open a lot of connections.  While running 500-node test, at some point (connecting the ovn-scale-325) the numerical value of a file descriptor surpasses 1024 and select() is unable to handle that throwing the exception:

```
2022-10-05 21:16:35,527 | ovsdbapp.backend.ovs_idl.vlog |INFO| ssl:192.16.1.74:6640: connecting...
2022-10-05 21:16:35,527 | ovsdbapp.backend.ovs_idl |ERROR| OVS database connection to Open_vSwitch failed with error: 'filedescriptor out of range in select()'. Verify that OVS and related services are available and that the relevant configuration options are correct.
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/ovsdbapp/backend/ovs_idl/__init__.py", line 109, in start_connection
    self.ovsdb_connection.start()
  File "/usr/local/lib/python3.10/site-packages/ovsdbapp/backend/ovs_idl/connection.py", line 83, in start
    idlutils.wait_for_change(self.idl, self.timeout)
  File "/usr/local/lib/python3.10/site-packages/ovsdbapp/backend/ovs_idl/idlutils.py", line 242, in wait_for_change
    ovs_poller.block()
  File "/usr/local/lib64/python3.10/site-packages/ovs/poller.py", line 231, in block
    events = self.poll.poll(self.timeout)
  File "/usr/local/lib64/python3.10/site-packages/ovs/poller.py", line 137, in poll
    rlist, wlist, xlist = select.select(self.rlist,
ValueError: filedescriptor out of range in select()
```

The way to avoid that is to replace the poller back to system-default select.poll like this:
```
from ovs import poller
import select
poller.SelectPoll = select.poll
```

This removes the limitation, since the poll() is correctly handling any number of file descriptors and not limited by their numerical values.  However, ovsdbapp is using multithreading, so I'm wondering what impact this change will have in runtime.
@otherwiseguy , maybe you can comment on this?

ovn-tester/ovn_utils.py
```diff
@@ -377,141 +518,128 @@ def acl_add(
...
-            cmd="ovn-sbctl --no-leader-only " + cmd,
-            stdout=stdout,
-            timeout=timeout,
+class OvnSbctl:
+    def __init__(self, sb, connection_string, inactivity_probe):
+        i = connection.OvsdbIdl.from_server(
+            connection_string, "OVN_Southbound"
         )
+        c = connection.Connection(i, inactivity_probe)
+        self.idl = SBIdl(c)
```
igsilya, line 633: It looks like this will monitor the whole SbDB.  I don't think we want that.  For example, we don't want to receive all the logical flows and spend a lot of time and memory on parsing and storing them.  We should subclass the connection like neutron does and only register tables that we're interested in.  In this case we only need a `Connection` and `Chassis` tables.

physical-deployments/physical-deployment.yml
```diff
@@ -4,6 +4,10 @@ internal-iface: eno1
 central-node:
   name: host02.mydomain.com
 
+tester-node:
+  name: host02.mydoman.com
+  ssh_key: id_rsa
```
igsilya, line 9: The full path is required here, so it should be better to have something like `/full/path/to/id_rsa` here.  We shouldn't use `/root/.ssh/id_rsa` in the example though to avoid accidental leak of the actual private key.

igsilya (changes requested):

do.sh
dceara: Ok, makes sense.  What would you think about adding a section in the readme documenting the steps that need to be done when changing the ovn-tester code?  It doesn't have to be very detailed I guess.

ovn-tester/ovn_utils.py
igsilya: Something like this, I guess (some re-naming is required):
```
@@ -612,6 +637,16 @@ class OvnNbctl:
         ):
             pass
 
+from ovsdbapp.backend.ovs_idl import idlutils
+class BaseOvnSbIdl(connection.OvsdbIdl):
+    schema = "OVN_Southbound"
+
+    @classmethod
+    def from_server(cls, connection_string):
+        helper = idlutils.get_schema_helper(connection_string, cls.schema)
+        helper.register_table('Chassis')
+        helper.register_table('Connection')
+        return cls(connection_string, helper)
+
 
 class SBIdl(sb_impl_idl.OvnSbApiIdlImpl, Backend):
     def __init__(self, connection):
@@ -626,9 +660,7 @@ class SBIdl(sb_impl_idl.OvnSbApiIdlImpl, Backend):
 
 class OvnSbctl:
     def __init__(self, sb, connection_string, inactivity_probe):
-        i = connection.OvsdbIdl.from_server(
-            connection_string, "OVN_Southbound"
-        )
+        i = BaseOvnSbIdl.from_server(connection_string)
         c = connection.Connection(i, inactivity_probe)
         self.idl = SBIdl(c)
 
```

physical-deployments/physical-deployment.yml
putnopvut: A full path is not actually required here. The reason I have it set to just `id_rsa` is because I generated an SSH key in the same directory as `do.sh`, and I used that key during testing.

However, I think it's a much better example to use a full path instead, so I will change it.

do.sh
putnopvut: Alternate solution number 4: Only run ovn-tester in the UTC timezone :)

putnopvut: I resolved this by removing timedelta logic from `latency.py`.

do.sh
putnopvut: I resolved this by moving the `cp` into `mine_data()` so that we don't exit if the HTML files are not present.

ovn-tester/ovn_utils.py
putnopvut: I used your suggested method for fixing this by switching from `select` to `poll`

ovn-tester/ovn_utils.py
putnopvut: I applied this fix. Thanks!

physical-deployments/physical-deployment.yml
putnopvut: I resolved this by using a full path in the config file.

putnopvut:
Commit fb170c8 adds gc debugging. I strongly suggest removing this commit before merging. However, while we are using this PR to do higher-scale tests, it may be useful to have this while debugging the latency spikes we see.

The reason I suggest removing this is because the garbage collector is EXTREMELY NOISY. It easily doubles or triples the size of `test-log` because it prints information all the time about garbage collection.

putnopvut:
Current status: This is not ready to be merged for two reasons:

1. @dceara's  note about adding information about ovn-tester installation to the README file.
2. @igsilya pointed out that process monitoring is not working as expected on the ovn-tester. The apparent reason is because the process name is `python3` rather than `ovn-tester`.

ovn-tester/ovn_tester.py
```diff
@@ -312,6 +351,17 @@ def run_base_cluster_bringup(ovn, bringup_cfg, global_cfg):
...
 
+def start_process_monitor(container):
+    log.info(f'Starting process monitor for {container}')
+    cmd = shlex.split(
+        f'bash -c "nohup python3 /tmp/process-monitor.py '
+        f'-s {container} '
+        f'-o /var/log/process-stats.json '
+        f'-x /tmp/process-monitor.exit"'
+    )
+    subprocess.Popen(cmd)
```
igsilya, line 362: Can we have a TesterNode class instead and not duplicate this function?
Another option is to start the process monitor from ansible, i.e. the place where we actually start that container.

.cirrus.yml
```diff
@@ -5,21 +5,23 @@ low_scale_task:
     image: family/fedora-cloud-35
     platform: linux
     memory: 8G
+    use_static_ip: true
```
igsilya, line 8: Do we actually need this?

.cirrus.yml
```diff
@@ -5,21 +5,23 @@ low_scale_task:
...
   env:
     DEPENDENCIES: git ansible podman podman-docker
-    PHYS_DEPLOYMENT: ${CIRRUS_WORKING_DIR}/physical-deployments/localhost.yml
+    PHYS_DEPLOYMENT: ${CIRRUS_WORKING_DIR}/physical-deployments/ci.yml
 
   runtime_cache:
     folder: runtime-cache
 
   configure_ssh_script:
+    - export IP_ADDR=$(ip route get 8.8.8.8 | head -1 | sed 's/.*src \([0-9\.]*\).*/\1/')
```
igsilya, line 18: Please, wrap this line.

.cirrus.yml
```diff
@@ -30,6 +32,8 @@ low_scale_task:
     - tar -xzf runtime-cache/runtime.tar.gz || true
 
   install_script:
+    - export IP_ADDR=$(ip route get 8.8.8.8 | head -1 | sed 's/.*src \([0-9\.]*\).*/\1/')
```
igsilya, line 35: And this.

do.sh
```diff
@@ -317,9 +347,32 @@ function mine_data() {
...
+        (( ip_index += 3 ))
+    else
+        (( ip_index += 1 ))
+    fi
+    n_relays=$(${ovn_fmn_get} ${test_file} cluster n_relays --default=0)
+    (( ip_index += ${n_relays} ))
+    tester_ip=$(${ovn_fmn_ip} ${node_net} ${node_ip} ${ip_index})
+    echo "${tester_ip}/${node_cidr}"
+}
+
```
igsilya, line 375: Can we make ovn-tester use the first IP and make the ovn-tester.py figure out IPs for the rest of the deployments?
Specifying defaults in two different places seems very error-prone.

ovn-fake-multinode-utils/playbooks/pull-ovn-tester.yml
```diff
@@ -0,0 +1,9 @@
+- name: Install ovn-tester
+  hosts: ovn_hosts
+  tasks:
+    - name: Pull latest containers
+      when: ovn_tester is defined
+      shell: |
+              set -e
+              docker pull {{ registry_node }}:5000/ovn/ovn-tester
+              docker tag {{ registry_node }}:5000/ovn/ovn-tester ovn/ovn-tester
```
igsilya, line 9: Indentation seems a bit off.
And why we need to tag the container here?  Should this be done after the build instead?

igsilya (commented):
Also, something is wrong with the test or a latency script as it reports following for the 250-node cluster-density:
```
min     : 10 ms
max     : 5311 ms
avg     : 1815 ms
95%     : 3966 ms
total   : 19836
failures: 22664
```
Number of failures doesn't make a lot of sense.

ovn-tester/ovn_tester.py
```diff
@@ -236,6 +250,23 @@ def read_config(config):
     return global_cfg, cluster_cfg, brex_cfg, bringup_cfg
 
 
+def setup_logging(global_cfg):
+    FORMAT = '%(asctime)s | %(name)-12s |%(levelname)s| %(message)s'
+    logging.basicConfig(stream=sys.stdout, level=logging.INFO, format=FORMAT)
```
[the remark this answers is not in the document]
putnopvut: Are you referring to the `__main__` on that first line? That's because we get our logger using `logging.getLogger(__name__)`. In imported modules, `__name__` is the name of the module. In the entry point python file, `__name__` evaluates to `__main__` . A fix for this would be to use the filename or the constant "ovn-tester" instead.

ovn-tester/ovn_tester.py
putnopvut: It's not exactly a duplicate of the other function, since this is not using ssh or `docker exec` like the other nodes and instead runs a process directly on the running container.

However, the suggestion to run it from ansible sounds good too.

physical-deployments/ci.yml
```diff
@@ -0,0 +1,12 @@
+registry-node: <ip>
+internal-iface: eno1
```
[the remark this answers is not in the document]
putnopvut: Sure, it's worth trying.

utils/latency.py
```diff
@@ -1,56 +1,40 @@
...
-if len(sys.argv) < 4:
-    print(
-        f'Usage {sys.argv[0]} "$(date +%z)" '
-        f'ovn-binding.log ovn-installed.log'
-    )
+if len(sys.argv) < 3:
+    print(f'Usage {sys.argv[0]} ' f'ovn-binding.log ovn-installed.log')
 
-tz = -int(sys.argv[1])
-delta = timedelta(hours=int(tz / 100), minutes=int(tz % 100))
```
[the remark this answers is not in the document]
putnopvut: Doesn't the time zone not matter so long as all containers are running the same time zone? The ovn-tester container uses the same base image as the other containers in the test, so it should be using the same time zone as the others. So whether they all use UTC or whether they all use UTC+3 or UTC-4, it shouldn't really matter.

.cirrus.yml
putnopvut: No, that's a holdover from before I was using introspection to determine the proper IP. According to Cirrus's documentation, you can set `use_static_ip: true` and you are supposed to have a guaranteed IP address assigned. However, in practice this did not work.

do.sh
putnopvut: I'll need to have another look at how ovn-fake-multinode's IP address assignment works. When doing this work, I did not think it would be possible because of inflexibilities in how ovn-fake-multinode assigns central and relay addresses. However, I may be able to cheat it by just bumping the starting address by 1 when starting central and relay containers in ovn-tester.

ovn-fake-multinode-utils/playbooks/pull-ovn-tester.yml
putnopvut: I'll double-check the indentation, but the comment about tagging after the build doesn't make sense to me. In the `install()` function in `do.sh`, we call:
````
install_ovn_tester
pull_ovn_tester
````
The first line builds the tester. The second calls the `pull-ovn-tester.yml` ansible playbook.

**Comment by putnopvut** [comment]
> Also, something is wrong with the test or a latency script as it reports following for the 250-node cluster-density:
> 
> ```
> min     : 10 ms
> max     : 5311 ms
> avg     : 1815 ms
> 95%     : 3966 ms
> total   : 19836
> failures: 22664
> ```
> 
> Number of failures doesn't make a lot of sense.

Can you link the test-log that resulted in this, please?

File: utils/latency.py (hunk 19 above)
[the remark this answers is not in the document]
Comment (igsilya) [reply utils/latency.py:13]: The problem is that OVS daemons (ovn-controller) logs always in UTC, regardless of the time zone of the system/container.  You can see the `Z` a.k.a 'Zulu time' in the ovn-controller log.  So, if the time zone configured in the container is not UTC, results will be skewed.

File: .cirrus.yml (hunk 12 above)
Comment (igsilya) [reply .cirrus.yml:8]: I guess, 'guaranteed IP address' is for how external services will see it, it doesn't mean that VM will have the same IP address, as that will obviously not work since they need to run more than one VM at the same time.

But they can nat everything to the same external IP, so you can add it to the allow list for your service, if CI needs to access that service, but you don't want to open it for the entire world.

File: ovn-tester/ovn_tester.py (hunk 11 above)
Comment (igsilya) [reply ovn-tester/ovn_tester.py:362]: I guess, SSHing to itself should not be problem.  But ansible way is also fine.

File: ovn-fake-multinode-utils/playbooks/pull-ovn-tester.yml (hunk 16 above)
Comment (putnopvut) [reply ovn-fake-multinode-utils/playbooks/pull-ovn-tester.yml:9]: I used the installation of ovn-multinode as an example here.  In `do.sh`, `install_ovn_fake_multinode` tags and pushes the `ovn-multi-node` image. Then `pull_ovn_fake_multinode` uses the `pull-fake-multinode.yml` playbook, on relevant hosts to pull and tag the ovn-multinode image.

To be honest, this struck me as odd, too, but since it's being done with ovn-multinode, and my focus is more on using ovsdbapp, I just mimicked the behavior with ovn-tester as well.

File: utils/latency.py (hunk 19 above)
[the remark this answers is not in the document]
Comment (putnopvut) [reply utils/latency.py:13]: Ah, OK, and since the ovn-installed times are pulled from ovn-controller logs, then that could potentially cause an issue. I'll add the timedelta calculation back in.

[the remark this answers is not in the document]
Comment (igsilya) [reply utils/latency.py:13]: There seems to be a way to subclass the logging formatter so ovn-tester will log in UTC.  Like this:
```
class UTCFormatter(logging.Formatter):
    converter = time.gmtime
```
Taken from here: https://docs.python.org/3/howto/logging-cookbook.html#formatting-times-using-utc-gmt-via-configuration

**Comment by putnopvut** [comment]
I fixed many of the issues pointed out in the PR. The remaining issues are:
* Update README to include details about re-installation of ovn-tester
* Get process-monitor working properly on ovn-tester node. As of now, it still is not collecting stats, and I have not figured out the magic incantation needed to get the ovn-tester process monitored properly.
* @igsilya's comment about tagging the image a second time is unresolved
* The odd latency calculations for the 250 node test are unresolved.

**Comment by putnopvut** [comment]
@igsilya I have good news and bad news about the odd latencies. The good news is that there is no bug in `latency.py`. The bad news is that lots of ovn-installed log messages are not present in the ovn-controller logs. ovn-scale-0 seems to be the only container where everything worked as expected. On all the other containers, either ports 10-161 or 11-161 have no message printed about being installed. The missing ovn-installed messages are why the failure list is so high. I have not looked into why those messages are missing from the log.

**Comment by igsilya** [comment]
> @igsilya I have good news and bad news about the odd latencies. The good news is that there is no bug in `latency.py`. The bad news is that lots of ovn-installed log messages are not present in the ovn-controller logs. ovn-scale-0 seems to be the only container where everything worked as expected. On all the other containers, either ports 10-161 or 11-161 have no message printed about being installed. The missing ovn-installed messages are why the failure list is so high. I have not looked into why those messages are missing from the log.

This is likely a side effect of my suggestion to actually add and delete ports in the startup phase of the cluster-density test that was implemented here: https://github.com/dceara/ovn-heater/pull/140 .
I would guess that we're removing these `build_ports` faster than controllers are able to claim them and that is causing the issue.  We should either not create them or wait for them to be claimed before removing.  Probably the former is the best option.
So, it's most likely not an issue of a current patch set.
CC: @dceara

**Comment by putnopvut** [comment]
Latest push updates the README and gets `process-monitor.py` working properly on the ovn-tester container. The only outstanding issue at this point is the re-tagging of the ovn-tester container. I haven't addressed this yet because I want to get some input from @dceara.

**Comment by dceara** [comment]
> Latest push updates the README and gets `process-monitor.py` working properly on the ovn-tester container. The only outstanding issue at this point is the re-tagging of the ovn-tester container. I haven't addressed this yet because I want to get some input from @dceara.

I don't think we need the re-tagging in any of the ansible playbooks.  "docker pull" should be enough.  But it's ok if you prefer to do this as a follow up, thanks!

**Review by dceara — COMMENTED** [review]
I had a closer look at all changes and they mostly look good to me.  I left a few comments though.  I think the only important comments are about the resulting html files and the load balancer related transactions.

Thanks!

File: ovn-tester/ovn_utils.py [hunk 20]
```diff
@@ -377,141 +561,144 @@ def acl_add(
...
     def lb_clear_vips(self, lb):
-        self.run(cmd=f'clear Load_Balancer {lb.uuid} vips')
+        self.idl.db_clear("Load_Balancer", lb.uuid, "vips").execute()
 
     def lb_add_to_routers(self, lb, routers):
-        cmd = ' -- '.join([f'lr-lb-add {r} {lb.uuid}' for r in routers])
-        self.run(cmd=cmd)
+        with self.idl.transaction(check_error=True) as txn:
+            for r in routers:
+                txn.add(self.idl.lr_lb_add(r, lb.uuid))
```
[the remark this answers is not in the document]
putnopvut: Yes, when used as a context manager (i.e. a `with` statement) the transaction executes once we exit the context manager.

do.sh
```diff
@@ -338,11 +379,14 @@ function run_test() {
     # Perform a fast cleanup by doing a minimal redeploy.
     init_ovn_fake_multinode
 
-    source ${rundir}/${ovn_heater_venv}/bin/activate
-    pushd ${out_dir}
+    tester_ip=$(get_tester_ip ${test_file})
+    if ! ansible-playbook ${ovn_fmn_playbooks}/run-tester.yml -i ${hosts_file} --extra-vars "test_file=${test_file} tester_ip=${tester_ip}" ; then
+         echo "-- Failed to set up test!"
```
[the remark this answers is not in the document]
putnopvut: I'm fine with making this change, but it would actually constitute a behavior change as compared to the master branch's behavior. Does it make sense to lump this behavior change in with this particular PR, or does it make sense to follow up with a separate one?

do.sh
```diff
@@ -338,11 +379,14 @@ function run_test() {
...
+    fi
 
-    if ! python -u ${ovn_tester}/ovn_tester.py $phys_deployment ${test_file} 2>&1 | tee ${ovn_tester_log_file}; then
-        echo "-- Failed to run test! Check logs at: $PWD/${ovn_tester_log_file}"
+    tester_host=$(${ovn_fmn_get} ${phys_deployment} tester-node name)
+    if ! ssh root@${tester_host} docker exec ovn-tester python3 -u /ovn-tester/ovn_tester.py /physical-deployment.yml /test-scenario.yml 2>&1 | tee ${out_dir}/test-log ; then
+        echo "-- Failed to run test. Check logs at: ${out_dir}/test-log"
```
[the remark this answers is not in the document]
putnopvut: Actually my comment above applies here. The master branch doesn't currently exit if running the test fails.

ovn-fake-multinode-utils/get-config-value.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
putnopvut: This is just a habit of mine. We can use either, but using python3 ensures that we'll use python 3 instead of 2. Hopefully nobody is defaulting to python 2 at this point though.

ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
```diff
@@ -11,6 +11,15 @@
     shell: |
       docker volume prune -f
 
+  - name: Stop tester container if running
+    when: ovn_tester is defined
+    ignore_errors: yes
+    shell: |
+      cd {{ ovn_fake_multinode_target_path }}/ovn-fake-multinode
+      ./ovs-docker del-ports br-ovn ovn-tester
+      ./ovs-docker del-ports br-ovn-ext ovn-tester
```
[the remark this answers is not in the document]
putnopvut: It couldn't hurt to remove this.

ovn-tester/ovn_utils.py
```diff
@@ -194,60 +240,199 @@ def unbind_vm_port(self, lport):
...
-        # FIXME: the SSH connection might have already been closed here..
-        # self.stop_daemon()
-        pass
+    def nb_has_completed(self, next_cfg):
+        if not self.wait_type:
+            return True
+        elif self.wait_type == "sb":
+            cur_cfg = self.api._nb.sb_cfg
+        else:  # self.wait_type == "hv":
+            cur_cfg = min(self.api._nb.sb_cfg, self.api._nb.hv_cfg)
```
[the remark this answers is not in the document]
putnopvut: I was about to mention about how python doesn't have issues with rollover like C. But since the DB server is likely limiting the value to LLONG_MAX, we need to deal with that in our python code as well.

And to answer the first question, I copied the behavior from `ovn-nbctl.c` : 
````
int64_t cur_cfg = (wait_type == NBCTL_WAIT_SB
                   ? nb->sb_cfg
                   : MIN(nb->sb_cfg, nb->hv_cfg));
````

File: do.sh [hunk 26]
```diff
@@ -321,9 +348,23 @@ function mine_data() {
...
+    deactivate
+
     popd
 }
 
+function get_tester_ip() {
+    local test_file=$1
+
+    # The tester gets the first IP address in the configured node_net.
+    node_net=$(${ovn_fmn_get} ${test_file} cluster node_net --default=192.16.0.0/16)
```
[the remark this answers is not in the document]
putnopvut: I was thinking about this, and there are a few ways that could work.
1. Use an environment variable to define default values and ensure this same environment variable is used both in the orchestrator and the tester container.
2. Define default values in some file, and refer to this file from `do.sh` and `ovn-tester`.
3. Write a python module similar to `ovn-fake-multinode-tools/get-config-value.py` and put it in `ovn-tester` and define defaults in that module.  Then `do.sh` and `ovn-tester` could both call into this module to get configuration values.
4. Use a template value in `ovn_tester.py` that can be filled in by `do.sh` at install-time.

Of these, I think 1 is clearly the worst. 2 is simple but also a bit clumsy. For me, it's a toss-up whether 3 or 4 is better. I think I would lean more towards option 3 myself. What do you think?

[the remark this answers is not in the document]
igsilya: Not sure if that is what you meant in 4, but one thing I had in mind previously was:
1. Make ovn-tester fail if at least one variable is not defined in the yml file, i.e. no defaults.
2. Make do.sh generate the full yml from a test scenario yml filling all the undefined values with defaults.
3. Run ovn-tester with a generated fully filled yml.

This will ensure that we do not have defaults defined in multiple places.  Will also be easier to track down undefined/misconfigured values if ovn-tester will just fail and not proceed using unexpected config values.

do.sh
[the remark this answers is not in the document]
dceara: Oops, sorry, you're right.  If the ovn_tester.py call fails we should gracefully report that and continue and collect logs/mine data.

do.sh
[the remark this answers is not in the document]
dceara: run-tester.yml is "infrastructure" so it should "never" fail, I think.  I agree with the ovn_tester.py part below.

do.sh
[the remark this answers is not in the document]
dceara: Sounds interesting! It also sounds like something we can do as a follow up in my opinion.  Shall we open an issue to track this?

ovn-fake-multinode-utils/get-config-value.py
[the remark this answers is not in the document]
dceara: If you feel like changing the other occurrences too that would be great.  Otherwise, no big deal, I can do it in a follow up PR.  Just let me know what you prefer, thanks!

ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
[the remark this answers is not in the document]
dceara: Ok, I'd remove it then.

ovn-tester/ovn_utils.py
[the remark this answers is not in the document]
dceara: Oh, oops, this was introduced to fix something I had reported :)
https://github.com/ovn-org/ovn/commit/cd3c0e1c2ab3a230965987e6c2be6cf8218226f6

It's to cover the corner case when there's no active chassis in the cluster.

For "dealing" with the overflow we can maybe check the value of `nb_cfg` incrementing it in `NBTransaction.pre_commit()`.

do.sh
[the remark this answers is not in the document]
putnopvut: @igsilya Sorry for the ambiguity on number 4. What I meant was that `ovn_tester.py` would have something like this in it:
````
DEFAULT_NODE_NET = {{ default_node_net }}
node_net = netaddr.IPNetwork(cluster_args.get('node_net', DEFAULT_NODE_NET))
````
We might also rename it to `ovn_tester.py.in` to make it clear that it can't be run as-is. Then, when running `do.sh install`, the `do.sh` code will populate `{{ default_node_net }}` with the proper default. This way, the default only is defined in `do.sh`.
It's not my favorite suggestion, but it's something I thought of.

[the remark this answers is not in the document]
dceara: @putnopvut Hmm, I'm not sure I like the idea of an `ovn_tester.py.in`.  I think I'd prefer @igsilya's suggestion of do.sh filling in all required input values in the yml we finally pass to ovn-tester.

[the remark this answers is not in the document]
putnopvut: Yes, that's why from the 4 suggestions I came up with, 3 was the one I preferred most. Mine and @igsilya's are more-or-less the same, except that in his, `do.sh` controls the defaults, and in mine, `ovn-tester` controls the defaults.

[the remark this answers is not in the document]
dceara: I see, your option "3" is also fine, thanks!

ovn-fake-multinode-utils/get-config-value.py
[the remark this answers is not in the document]
putnopvut: I elected to be consistent and just use "python" in the latest version. We can do a project-wide change to "pyhon3" as a follow-up potentially.

ovn-tester/ovn_utils.py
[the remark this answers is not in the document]
putnopvut: I just pushed new changes and realized I did not address this. I'll push another update in just a bit.

putnopvut:
As of the latest update, the only outstanding comment is https://github.com/dceara/ovn-heater/pull/136#discussion_r1008006173 . Should I attempt to do this as part of this PR or save it for a follow-up?

dceara:
> As of the latest update, the only outstanding comment is [#136 (comment)](https://github.com/dceara/ovn-heater/pull/136#discussion_r1008006173) . Should I attempt to do this as part of this PR or save it for a follow-up?

Follow up sounds good to me, let's try to get this PR in. It looks almost ready to me. I'll have another look tomorrow.

putnopvut:
> > As of the latest update, the only outstanding comment is [#136 (comment)](https://github.com/dceara/ovn-heater/pull/136#discussion_r1008006173) . Should I attempt to do this as part of this PR or save it for a follow-up?
> 
> Follow up sounds good to me, let's try to get this PR in. It looks almost ready to me. I'll have another look tomorrow.

OK nice. As of this moment, this PR has 36 commits associated with it, because of the many individual commits I did to address PR issues. Is it OK to merge these as-is or should I squash them into a more reasonable series of commits?

dceara:
> OK nice. -As of this moment, this PR has 36 commits associated with it, because of the many individual commits I did to address PR issues. Is it OK to merge these as-is or should I squash them into a more reasonable series of commits?

I think the current series of commits is reasonable enough.  I'd maybe squash [Address PR comments.](https://github.com/dceara/ovn-heater/pull/136/commits/f4a4ac62824726686cc14961fe6986f0f6d2f496), [Address some more PR comments:](https://github.com/dceara/ovn-heater/pull/136/commits/a1af949790b8deefcb276d9e374f16617bece150) and [Fix pep8 issues.](https://github.com/dceara/ovn-heater/pull/136/commits/b0d9ade92919f12cb23dc36742ca681540bae753) into the appropriate commits.  The rest is fine by me, thanks!

do.sh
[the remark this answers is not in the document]
putnopvut: I've actually implemented @igsilya's idea at https://github.com/putnopvut/ovn-heater/tree/default_dumper . I was hoping to be able to open a PR here, but github does not allow using a PR as the basis for another PR. Therefore, when I created the PR, it was way larger and more confusing than necessary. I'll hold off and post that PR once this one is merged.

dceara:
I mangled the commits a bit to include the review incremental updates in the correct commits.  Apart from that I didn't change anything.  I went ahead and pushed this to the main branch.  Thanks!

PR #140, opened 2022-10-04
Realistic startup.

dceara:
Make the startup stages for cluster-density and density-heavy provision things in the same way as the "main" test stage does.

### DIFF
--- a/ovn-tester/tests/cluster_density.py
+++ b/ovn-tester/tests/cluster_density.py
@@ -9,10 +9,11 @@ DENSITY_N_BUILD_PODS = 6
 DENSITY_N_PODS = 4
 DENSITY_N_TOT_PODS = DENSITY_N_BUILD_PODS + DENSITY_N_PODS
 
+# In ClusterDensity.run_iteration() we assume at least 4 different pods
+# can be used as backends.
+assert DENSITY_N_PODS >= 4
 
-ClusterDensityCfg = namedtuple(
-    'ClusterDensityCfg', ['n_runs', 'batch', 'n_startup']
-)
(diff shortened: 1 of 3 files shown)


dceara:
CC: @igsilya

ovn-tester/tests/cluster_density.py
```diff
@@ -23,72 +27,57 @@ def __init__(self, config, central_node, worker_nodes, global_cfg):
...
-                    # add VIPs and backends to namespace load-balancer
-                    ns.provision_vips_to_load_balancers(
-                        [ports[0:2], ports[2:3], ports[3:4]]
-                    )
+                # Create DENSITY_N_BUILD_PODS short lived "build" pods.
+                build_ports = ovn.provision_ports(
+                    DENSITY_N_BUILD_PODS, passive=False
+                )
+                ns.add_ports(build_ports)
+                ovn.ping_ports(build_ports)
```
igsilya, line 71: Should all that also be part of the iteration?  I'm guessing the port creation/deletion churn might be a valid part of the test.  It will be reflected in the ovn-installed latency at least.  Not sure.
If we'll go that path, we can move everything into the iteration function and have maybe an extra argument to mark if we need to execute a ping test or not.

dceara: That's a fair point, I'll do it like that, thanks for the review!

dceara: I updated the code as discussed and I did the same thing for density-heavy (included the ping in the run_iteration function).

igsilya (approved):
Didn't test, but the code looks OK.  Thanks!

PR #142, opened 2022-10-26
cluster_density: Skip build pods in the startup stage.

dceara:
We don't wait for them to be up so it's possible that the port gets added and removed from the NB before ovn-controller has the chance to claim them.  That will mess up the 'ovn-installed' latency calculation as we expect all LSPs to be claimed eventually.

Reported-at: https://github.com/dceara/ovn-heater/pull/136#issuecomment-1276749002


dceara:
CC: @igsilya

igsilya (approved):
Didn't test, but LGTM.

PR #144, opened 2022-11-04
Centralize YAML Defaults

putnopvut:
In the pull request to use ovsdbapp [1] the issue [2] was raised that both `do.sh` and `ovn_tester.py` had duplicated a default value. It means that if that default were to change, we would need to update it in multiple places. This change implements the suggestion made by @igsilya to translate the test scenario YAML file into a YAML file with all defaults filled in.

Supposedly this change applies to the main branch of ovn-heater as-is, but I developed this in a branch based on PR 136, so it would be best if this went in after the ovsdbapp change was made.

[1] https://github.com/dceara/ovn-heater/pull/136
[2] https://github.com/dceara/ovn-heater/pull/136#discussion_r1008006173

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,6 @@ low_scale_task:
     image: family/fedora-cloud-35
     platform: linux
     memory: 8G
-    use_static_ip: true
 
   env:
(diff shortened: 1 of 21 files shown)


putnopvut:
Ah crap.

PR #145, opened 2022-11-15
config: Centralize default YAML.

putnopvut:
Test scenario configuration is fairly lean since all configuration values have some default value. In the good old days, the default values for configuration items could be dictated solely in ovn-tester.

With the containerization of the tester, now do.sh sometimes needs to use configuration values as well. This means that do.sh sometimes also needs to be able to use the default values. However, do.sh has no idea what these default values are, so they end up getting duplicated by developers.

This change seeks to solve this problem. A new python script is now the arbiter of default values. When running a test, do.sh will take the test configuration, run it through this script, and generate a yaml file that has all default values filled in. This is then passed to ovn-tester as the configuration file. ovn-tester can now accept the configuration file as-is and not have to try to fill in any default values.

### DIFF
--- a/ovn-fake-multinode-utils/translate_yaml.py
+++ b/ovn-fake-multinode-utils/translate_yaml.py
@@ -1,0 +1,196 @@
+#!/usr/bin/env python
+
+from pathlib import Path
+from dataclasses import dataclass
+from typing import Dict
+import yaml
+import netaddr
+import sys
+
+
+def load_yaml(orig_yaml_file_name):
+    with open(orig_yaml_file_name, "r") as orig_yaml_file:
+        return yaml.safe_load(orig_yaml_file)
+
+
+@dataclass
+class GlobalConfig:
+    """This contains all "global" level configuration options and their
+    default values. If you want to add a new global level option then it needs
+    to be listed here and have its type and default value specified."""
+
+    log_cmds: bool = False
+    cleanup: bool = False
+    run_ipv4: bool = True
+    run_ipv6: bool = False
+
+
+def calculate_node_remotes(
+    node_net: str, clustered_db: bool, n_relays: int, enable_ssl: bool
+) -> str:
+    net = netaddr.IPNetwork(node_net)
+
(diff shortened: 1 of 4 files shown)


dceara:
Sorry @putnopvut for lack of reviews on this one.  It still seems to apply cleanly.  I triggered a re-run of the cirrusCI here: https://cirrus-ci.com/task/6388560614391808

The changes look OK to me. @igsilya what do you think?

igsilya:
@dceara  I didn't test, but looks OK on a quick glance.
We may want to rebase this PR, because simple re-run of checks doesn't rebase things, it just runs using the same old branch that might not have some recent CI changes.

ovn-fake-multinode-utils/translate_yaml.py
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env python
+
+from pathlib import Path
+from dataclasses import dataclass
+from typing import Dict
+import yaml
```
igsilya, line 6: This requires installation of `python3-pyyaml` on the orchestrator in the `./do.sh` script.  `install-dependencies.yml` playbook installs it on all other nodes, but it is not executed on orchestrator if orchestrator and the tester are not the same node.

igsilya (changes requested):
I tested the change and it works fine, except for the missing pyyaml dependency that I commented on.  Code looks fine otherwise.

putnopvut:
I just pushed new changes. The main changes are:
- I addressed Ilya's comment about adding pyyaml on the orchestrator. I added pyyaml to the requirements.txt file, which is installed in the local venv by do.sh
- I changed some of the checks in translate_yaml.py so that we now explicitly check calculated values against "None" instead of using "not". This way, if someone purposely wants to have an empty VIP list, for instance, they can do that.
- I updated a comment in the GlobalConfig dataclass where it erroneously said to add cluster level options there.

dceara:
Thanks @putnopvut and @igsilya!

PR #150, opened 2023-02-15
ovn-tester: Disable neighbor_responder flows for LB VIPs on LR.

igsilya:
ovn-kubernetes disables this option since:
  https://github.com/ovn-org/ovn-kubernetes/pull/3400

### DIFF
--- a/ovn-tester/ovn_utils.py
+++ b/ovn-tester/ovn_utils.py
@@ -608,6 +608,7 @@ class OvnNbctl:
             "Load_Balancer",
             name=lb_name,
             protocol=protocol,
+            options={"neighbor_responder": "none"},
             get_func=partial(self.idl.lb_get, lb_name),
         )
         return LoadBalancer(name=lb_name, uuid=uuid)


ovn-tester/ovn_utils.py
```diff
@@ -610,6 +610,9 @@ def create_lb(self, name, protocol):
             protocol=protocol,
             get_func=partial(self.idl.lb_get, lb_name),
         )
+        self.idl.db_set(
+            "Load_Balancer", uuid, ("options", {"neighbor_responder": "none"})
+        ).execute()
```
dceara, line 615: A bit of a nit pick but would it be possible to set the option in the same transaction as the LB creation?

igsilya: TBH, I got a bit lost in python trying to figure out how to pass parameters into `self.db_create_transaction`.  I'll try one more time.

igsilya: OK.  Updated version seems to work fine.

dceara (approved):
Looks good to me, thanks!

PR #154, opened 2023-03-03
install-fake-multinode.yml: Copy ovn-fake-multinode from the orchestrator.

igsilya:
Instead of cloning the same repo from each physical host it can be copied from the orchestrator node, where we already have it.

Simultaneous git clone from multiple physical nodes frequently results in GitHub connection timeouts.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,7 +7,8 @@ low_scale_task:
     memory: 8G
 
   env:
-    DEPENDENCIES: git ansible podman podman-docker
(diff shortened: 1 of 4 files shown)


igsilya:
There is a problem if the orchestrator is part of the deployment.  Playbook removes the folder before copying.  Need to think how to resolve that.

igsilya:
'synchronize' can handle the sync of the directory to itself, no need to delete it.  Updated PR with that change.

PR #157, opened 2023-03-09
Introduce initial support for multiple distributions.

fnordahl:
Please review commit by commit.

Signed-off-by: Frode Nordahl <frode.nordahl@canonical.com>

### DIFF
--- a/do.sh
+++ b/do.sh
@@ -56,14 +56,9 @@ function install_deps() {
     ansible-playbook ${ovn_fmn_playbooks}/install-dependencies.yml -i ${hosts_file}
 
     echo "-- Installing local dependencies"
-    if yum install docker-ce --nobest -y
-    then
-        systemctl start docker
-    else
-        yum install -y podman podman-docker
-    fi
     yum install redhat-lsb-core datamash \
         python3-pip python3-virtualenv python3 python3-devel python-virtualenv \
+        podman podman-docker \
         --skip-broken -y
     [ -e /usr/bin/pip ] || ln -sf /usr/bin/pip3 /usr/bin/pip
 
--- a/do.sh
+++ b/do.sh
@@ -34,9 +34,43 @@ ovn_tester=${topdir}/ovn-tester
 EXTRA_OPTIMIZE=${EXTRA_OPTIMIZE:-no}
 USE_OVSDB_ETCD=${USE_OVSDB_ETCD:-no}
 
+# We want values from both the `ID` and `ID_LIKE` fields to ensure successful
+# categorization.  The shell will happily accept both spaces and newlines as
+# separators:
+# https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_05
+DISTRO_IDS=$(awk -F= '/^ID/{print$2}' /etc/os-release | tr -d '"')
+
+function is_rpm_based() {
+    for id in $DISTRO_IDS; do
+        case $id in
+        centos* | rhel* | fedora*)
+            true
+            return
+        ;;
+        esac
+    done
+    false
+}
+
+function is_deb_based() {
+    for id in $DISTRO_IDS; do
+        case $id in
+        debian* | ubuntu*)
+            true
+            return
+            ;;
+        esac
+    done
+    false
+}
+
 function die() {
     echo $1
     exit 1
+}
+
+function die_distro() {
+    die "Unable to determine distro type, rpm- and deb-based are supported."
 }
 
 function generate() {
@@ -51,17 +85,29 @@ function generate() {
     cp ${ovn_fmn_utils}/scripts/perf.sh ${log_perf_file}
 }
 
-function install_deps() {
-    echo "-- Installing dependencies on all nodes"
-    ansible-playbook ${ovn_fmn_playbooks}/install-dependencies.yml -i ${hosts_file}
-
+function install_deps_local_rpm() {
     echo "-- Installing local dependencies"
     yum install redhat-lsb-core datamash \
-        python3-pip python3-virtualenv python3 python3-devel python-virtualenv \
-        podman podman-docker \
+        python3-pip python3-virtualenv python3 python3-devel \
+        python-virtualenv podman podman-docker \
         --skip-broken -y
     [ -e /usr/bin/pip ] || ln -sf /usr/bin/pip3 /usr/bin/pip
 
+}
+
+function install_deps_local_deb() {
+    echo "-- Installing local dependencies"
+    apt -y install datamash podman podman-docker python3-pip \
+           python3-virtualenv python3 python3-all-dev python3-virtualenv
+}
+
+function install_deps_remote() {
+    echo "-- Installing dependencies on all nodes"
+    ansible-playbook ${ovn_fmn_playbooks}/install-dependencies.yml \
+        -i ${hosts_file}
+}
+
+function run_registry() {
     containers=$(docker ps --all --filter='name=(ovn|registry)' \
                         | grep -v "CONTAINER ID" | awk '{print $1}' || true)
     for container_name in $containers
@@ -72,10 +118,10 @@ function install_deps() {
     [ -d /var/lib/registry ] || mkdir /var/lib/registry -p
     docker run --privileged -d --name registry -p 5000:5000 \
           -v /var/lib/registry:/var/lib/registry --restart=always docker.io/library/registry:2
-        cp /etc/containers/registries.conf /etc/containers/registries.conf.bak
-        cat > /etc/containers/registries.conf << EOF
-[registries.search]
-registries = ['registry.access.redhat.com', 'registry.redhat.io']
+
+    # This is requried on the orchestrator for local image build/push to work
+    cp /etc/containers/registries.conf /etc/containers/registries.conf.bak
+    cat > /etc/containers/registries.conf << EOF
 [registries.insecure]
 registries = ['localhost:5000']
 [registries.block]
@@ -257,7 +303,17 @@ function pull_ovn_tester() {
 
 function install() {
     pushd ${rundir}
-    install_deps
+    if is_rpm_based
+    then
+        install_deps_local_rpm
+    elif is_deb_based
+    then
+        install_deps_local_deb
+    else
+        die_distro
+    fi
+    install_deps_remote
+    run_registry
     install_venv
     configure_docker
     install_ovn_fake_multinode
--- a/ovn-fake-multinode-utils/get-config-value.py
+++ b/ovn-fake-multinode-utils/get-config-value.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 import argparse
 import yaml
 
--- a/ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
+++ b/ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
@@ -3,9 +3,16 @@
   become: true
   tasks:
   - name: Start openvswitch
-    systemd:
+    ansible.builtin.systemd:
       name: openvswitch
       state: started
+    when: ansible_os_family == "RedHat"
+
+  - name: Start openvswitch
+    ansible.builtin.systemd:
+      name: openvswitch-switch
+      state: started
+    when: ansible_os_family == "Debian"
 
   - name: Delete old docker volumes
     shell: |
@@ -34,7 +41,18 @@
     shell: |
       sed -i '/^MaxSessions /d' /etc/ssh/sshd_config
       echo "MaxSessions 1024" >> /etc/ssh/sshd_config
-      systemctl reload sshd.service
+
+  - name: Reload SSH service
+    ansible.builtin.systemd:
+      name: sshd
+      state: reloaded
+    when: ansible_os_family == "RedHat"
+
+  - name: Reload SSH service
+    ansible.builtin.systemd:
+      name: ssh
+      state: reloaded
+    when: ansible_os_family == "Debian"
 
   - name: Ensure the neighbor cache size on central nodes
     when: ovn_central is defined
--- a/ovn-fake-multinode-utils/playbooks/install-dependencies.yml
+++ b/ovn-fake-multinode-utils/playbooks/install-dependencies.yml
@@ -2,6 +2,24 @@
   hosts: ovn_hosts
   tasks:
     - name: Install required packages
-      shell: yum install git gcc openvswitch python3-pyyaml python3-virtualenv python3-devel containers-common --skip-broken -y
+      shell: |
+        yum install git gcc openvswitch python3-pyyaml python3-virtualenv \
+            python3-devel containers-common --skip-broken -y
+      when: ansible_os_family == "RedHat"
+    - name: Install required packages
+      ansible.builtin.package:
+        name:
+          - git
+          - gcc
+          - openvswitch-switch
+          - python3-yaml
+          - python3-virtualenv
+          - python3-all-dev
+        state: present
+      when: ansible_os_family == "Debian"
     - name: Install container command
-      shell: yum install podman podman-docker -y
+      ansible.builtin.package:
+        name:
+          - podman
+          - podman-docker
+        state: present
--- a/ovn-fake-multinode-utils/playbooks/install-dependencies.yml
+++ b/ovn-fake-multinode-utils/playbooks/install-dependencies.yml
@@ -2,10 +2,28 @@
   hosts: ovn_hosts
   tasks:
     - name: Install required packages
-      shell: |
-        yum install git gcc openvswitch python3-pyyaml python3-virtualenv \
-            python3-devel containers-common --skip-broken -y
+      ansible.builtin.package:
+        name:
+          - git
+          - gcc
+          - python3-pyyaml
+          - python3-virtualenv
+          - python3-devel
+          - containers-common
+        state: present
       when: ansible_os_family == "RedHat"
+
+    # 'skip-broken' might not work with specific ansible versions
+    # so ignore_errors for packages that might not be there.
+    # https://github.com/ansible/ansible/commit/6bcb494f8306
+    - name: Install optional packages
+      ansible.builtin.package:
+        name:
+          - openvswitch
+        state: present
+      ignore_errors: true
+      when: ansible_os_family == "RedHat"
+
     - name: Install required packages
       ansible.builtin.package:
         name:
@@ -17,6 +35,7 @@
           - python3-all-dev
         state: present
       when: ansible_os_family == "Debian"
+
     - name: Install container command
       ansible.builtin.package:
         name:
--- a/ovn-fake-multinode-utils/playbooks/install-dependencies.yml
+++ b/ovn-fake-multinode-utils/playbooks/install-dependencies.yml
@@ -4,4 +4,4 @@
     - name: Install required packages
       shell: yum install git gcc openvswitch python3-pyyaml python3-virtualenv python3-devel containers-common --skip-broken -y
     - name: Install container command
-      shell: yum install docker-ce --nobest -y || yum install docker -y || yum install podman podman-docker -y
+      shell: yum install podman podman-docker -y
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,14 +1,18 @@
 low_scale_task:
 
   compute_engine_instance:
-    image_project: fedora-cloud
-    image: family/fedora-cloud-35
+    matrix:
+      - image_project: fedora-cloud
+        image: family/fedora-cloud-35
+      - image_project: ubuntu-os-cloud
+        image: family/ubuntu-2204-lts
     platform: linux
     memory: 8G
+    disk: 20
 
   env:
-    DEPENDENCIES: git ansible ansible-collection-ansible-posix
-                  podman podman-docker
+    DEPENDENCIES: git ansible podman podman-docker
+    FEDORA_DEP: ansible-collection-ansible-posix
     PHYS_DEPLOYMENT: ${CIRRUS_WORKING_DIR}/physical-deployments/ci.yml
 
   runtime_cache:
@@ -23,7 +27,12 @@ low_scale_task:
     - ssh root@$(hostname) -v echo Hello
 
   install_dependencies_script:
-    - dnf install -y ${DEPENDENCIES}
+    - 'if [ $(which dnf) ]; then
+        dnf install -y ${DEPENDENCIES} ${FEDORA_DEP};
+       fi'
+    - 'if [ $(which apt) ]; then
+        apt update && apt install -y ${DEPENDENCIES};
+       fi'
 
   unpack_caches_script:
     - mkdir -p runtime runtime-cache
--- a/README.md
+++ b/README.md
@@ -66,29 +66,56 @@ not resolve to a unique host.
 ## Minimal requirements on the ORCHESTRATOR node (tested on Fedora 32)
 
 ### Install required packages:
+
+#### RPM-based
 ```
 dnf install -y git ansible ansible-collection-ansible-posix
 ```
 
+#### DEB-based
+```
+sudo apt -y install ansible
+```
+
 ## Minimal requirements on the TESTER node (tested on Fedora 36)
 
-### Make docker work with Fedora 32+ (disable cgroup hierarchy):
+### Make docker work with nested containers (disable cgroup hierarchy):
 
+#### RPM-based Fedora 32+
 ```
 dnf install -y grubby
 grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
 reboot
 ```
+
+#### DEB-based
+Edit /etc/default/grub and add `systemd.unified_cgroup_hierarchy=0` at the
+end of the `GRUB_CMDLINE_LINUX_DEFAULT` variable.
+
+```
+sudo update-grub
+sudo reboot
+````
 
 ## Minimal requirements on the OVN-CENTRAL and OVN-WORKER-NODEs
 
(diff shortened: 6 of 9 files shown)


igsilya:
Hi @fnordahl .  Thanks for the PR!
I didn't go too deep into review, but one thing that is needed here, I think, is CI.  I spent some time with cirrus and the following change seems to work (ubuntu images have less available disk space for some reason):
```
diff --git a/.cirrus.yml b/.cirrus.yml
index f24475b..69602d7 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,14 +1,19 @@
 low_scale_task:
 
   compute_engine_instance:
-    image_project: fedora-cloud
-    image: family/fedora-cloud-35
+    matrix:
+      - image_project: fedora-cloud
+        image: family/fedora-cloud-35
+      - image_project: ubuntu-os-cloud
+        image: family/ubuntu-2204-lts
     platform: linux
     memory: 8G
+    disk: 20
 
   env:
-    DEPENDENCIES: git ansible ansible-collection-ansible-posix
-                  podman podman-docker
+    DEPENDENCIES: git ansible
+    FEDORA_DEP: podman podman-docker ansible-collection-ansible-posix
+    UBUNTU_DEP: docker.io
     PHYS_DEPLOYMENT: ${CIRRUS_WORKING_DIR}/physical-deployments/ci.yml
 
   runtime_cache:
@@ -23,7 +28,12 @@ low_scale_task:
     - ssh root@$(hostname) -v echo Hello
 
   install_dependencies_script:
-    - dnf install -y ${DEPENDENCIES}
+    - 'if [ $(which dnf) ]; then
+        dnf install -y ${DEPENDENCIES} ${FEDORA_DEP};
+       fi'
+    - 'if [ $(which apt) ]; then
+        apt update && apt install -y ${DEPENDENCIES} ${UBUNTU_DEP};
+       fi'
 
   unpack_caches_script:
     - mkdir -p runtime runtime-cache
```

Would be great if you can include something like that, as missing dependencies are hard to track in ovn-heater.

On the other note, your current implementation seem to require docker, while for rpm-based distros we allow using podman as well.   Do you think it makes sense to allow podman for debian-based systems as well?
We also had a lot of issues with docker in the past (mainly a requirement to be logged in and some unfortunate deamon hangs), so the question of supporting only podman and run s/docker/podman/ on a codebase was brought up several times in the past.
Do you think it makes sense to support only podman?  Most of major distributions seem to package it (I didn't check debian).

Also, if I install podman in the .cirrus.yml for ubuntu I'm getting errors later from ./do.sh install, likely because podman is getting replaced by docker.io in the process and something goes wrong:
```
TASK [Pull latest containers] **************************************************
fatal: [cirrus-task-6259513473368064]: FAILED! => {"changed": true, "cmd": "set -e\ndocker pull cirrus-task-6259513473368064:5000/ovn/ovn-multi-node\ndocker tag cirrus-task-6259513473368064:5000/ovn/ovn-multi-node ovn/ovn-multi-node\n", "delta": "0:00:00.050686", "end": "2023-03-10 11:56:42.841776", "msg": "non-zero return code", "rc": 1, "start": "2023-03-10 11:56:42.791090", "stderr": "Error response from daemon: Get \"https://cirrus-task-6259513473368064:5000/v2/\": http: server gave HTTP response to HTTPS client", "stderr_lines": ["Error response from daemon: Get \"https://cirrus-task-6259513473368064:5000/v2/\": http: server gave HTTP response to HTTPS client"], "stdout": "Using default tag: latest", "stdout_lines": ["Using default tag: latest"]}
```

fnordahl:
Thank you for the review @igsilya and thanks a lot for contributing the Cirrus CI recipe, I included it verbatim in the PR with appropriate attribution etc. Making use of the `podman` and `podman-docker` packages appear to work fine, so happy to go with that, change incorporated in the updated version of the PR.

Cheers!

dceara:
> Thank you for the review @igsilya and thanks a lot for contributing the Cirrus CI recipe, I included it verbatim in the PR with appropriate attribution etc.

@fnordahl according to the [contributing guidelines](https://github.com/dceara/ovn-heater/blob/main/CONTRIBUTING.md#tags) (which I just pushed last week :D) all co-authors must sign off (like in ovs/ovn).  @igsilya are you ok with me or @fnordahl adding your `Signed-off-by` to the last commit?

igsilya:
> > Thank you for the review @igsilya and thanks a lot for contributing the Cirrus CI recipe, I included it verbatim in the PR with appropriate attribution etc.
> 
> @fnordahl according to the [contributing guidelines](https://github.com/dceara/ovn-heater/blob/main/CONTRIBUTING.md#tags) (which I just pushed last week :D) all co-authors must sign off (like in ovs/ovn). @igsilya are you ok with me or @fnordahl adding your `Signed-off-by` to the last commit?

Sure.  Though, I'd move the `podman` and `podman-docker` to `DEPENDENCIES` and removed the `UBUNTU_DEP` for a time being since we don't really need it.

do.sh
```diff
@@ -34,11 +34,39 @@ ovn_tester=${topdir}/ovn-tester
 EXTRA_OPTIMIZE=${EXTRA_OPTIMIZE:-no}
 USE_OVSDB_ETCD=${USE_OVSDB_ETCD:-no}
 
+DISTRO_IDS=$(awk -F= '/^ID/{print$2}' /etc/os-release)
```
dceara, line 37: Unfortunately distros don't make this easy.

This works on Fedora and Ubuntu but on my RHEL8 `/etc/os-release` I have:
```
ID="rhel"
ID_LIKE="fedora"
```
Which makes the above compute:
```
$ echo $DISTRO_IDS
"rhel" "fedora"
```
I changed it to
```
DISTRO_IDS=$(awk -F= '/^ID=/{print$2}' /etc/os-release | tr -d '"')
```
and this seemed to do the right thing for both Fedora and RHEL.

do.sh
```diff
@@ -34,11 +34,39 @@ ovn_tester=${topdir}/ovn-tester
...
+    esac
+}
+
 function die() {
     echo $1
     exit 1
 }
 
+function die_distro() {
+    die "Unable to determine distribution type."
```
dceara, line 67: Should we say here that we currently only detect rpm-based and deb-based distros?

do.sh
```diff
@@ -53,19 +81,37 @@ function generate() {
...
+        die_distro
+    fi
+    ansible-playbook ${ovn_fmn_playbooks}/${_install_dep_playbook} -i ${hosts_file}
 
     echo "-- Installing local dependencies"
-    if yum install docker-ce --nobest -y
+    if is_rpm_based
     then
-        systemctl start docker
+        if yum install docker-ce --nobest -y
```
dceara, line 99: I know it was like this before your change but we (RedHat) don't use docker-ce anymore downstream.  We use podman downstream (and in CirrusCI).  I wonder if we should just remove docker-ce all together and simplify things..

do.sh
```diff
@@ -53,19 +81,37 @@ function generate() {
...
+        yum install redhat-lsb-core datamash \
+            python3-pip python3-virtualenv python3 python3-devel python-virtualenv \
+            --skip-broken -y
+        [ -e /usr/bin/pip ] || ln -sf /usr/bin/pip3 /usr/bin/pip
+    elif is_deb_based
+    then
+        apt -y install datamash podman podman-docker python3-pip python3-virtualenv python3 python3-all-dev python3-virtualenv
     else
-        yum install -y podman podman-docker
+        die_distro
```
dceara, line 113: Instead of duplicating the `die_distro` check does it make sense to just do it once, in the beginning of the `install()` function?

do.sh
```diff
@@ -77,6 +123,8 @@ function install_deps() {
     [ -d /var/lib/registry ] || mkdir /var/lib/registry -p
     docker run --privileged -d --name registry -p 5000:5000 \
           -v /var/lib/registry:/var/lib/registry --restart=always docker.io/library/registry:2
+    if [ -d /etc/containers ]
```
dceara, line 126: Do we actually need this at all?  We don't pull anything on the orchestrator node so can this whole block go away?  Of course, not a problem of this commit.

ovn-fake-multinode-utils/get-config-value.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
dceara, line 1: `ovn-fake-multinode-utils/translate_yaml.py` was recently added.  We should change python to python3 there too I guess.

dceara (changes requested):
Thanks for the PR, looks good!  I just have a few (relatively minor) comments.

do.sh
fnordahl: :+1:

do.sh
fnordahl: In order to do this check once, we would need to change the structure of the program, i.e. we would need to install local deps first and then run the ansible playbook for other nodes afterwards, but I guess that's fine too

do.sh
fnordahl: Sure, I'll add a separate commit for that

do.sh
fnordahl: :+1:

do.sh
fnordahl: Fixed in https://github.com/dceara/ovn-heater/pull/157/commits/d92932f6b4eae36dbd3a8f4dff61c0c709086b4b

do.sh
fnordahl: Done

ovn-fake-multinode-utils/get-config-value.py
fnordahl: Done

do.sh
fnordahl: Fixed in https://github.com/dceara/ovn-heater/pull/157/commits/c7af0c657747450952e8eb495cad8116fbc80f52

do.sh
```diff
@@ -84,15 +84,10 @@ function install_deps() {
     local _install_dep_playbook
     if is_rpm_based
     then
-        if yum install docker-ce --nobest -y
-        then
-            systemctl start docker
-        else
-            yum install -y podman podman-docker
-        fi
```
igsilya: I guess, we should also remove an attempt to install docker-ce from the `ovn-fake-multinode-utils/playbooks/install-dependencies.yml`.

fnordahl: :+1:

do.sh
```diff
@@ -34,11 +34,39 @@ ovn_tester=${topdir}/ovn-tester
 EXTRA_OPTIMIZE=${EXTRA_OPTIMIZE:-no}
 USE_OVSDB_ETCD=${USE_OVSDB_ETCD:-no}
 
+DISTRO_IDS=$(awk -F= '/^ID/{print$2}' /etc/os-release | tr -d '"')
```
dceara, line 37: This works now but still sets `DISTRO_IDS` to `rhel fedora` on my RHEL machine.

Nit: I think I'd still change this to:
 ```
DISTRO_IDS=$(awk -F= '/^ID=/{print$2}' /etc/os-release | tr -d '"')
```

But I'll leave it up to you, it's OK as is too.

File: do.sh (hunk 4 above)
Comment (dceara) [reply do.sh:100]: Hmm, I'm sorry, I think my initial comment wasn't clear.  I actually meant this:
```
diff --git a/do.sh b/do.sh
index 6f00c08..b237600 100755
--- a/do.sh
+++ b/do.sh
@@ -303,6 +303,7 @@ function pull_ovn_tester() {
 }
 
 function install() {
+    ! is_rpm_based && ! is_deb_based && die_distro
     pushd ${rundir}
     install_deps
     install_venv
```

On a second thought, isn't it cleaner if we just add per distro `install_deps` functions.  E.g.:
```
function install() {
    pushd ${rundir}
    if is_rpm_based
    then
        install_deps_rpm
    elif is_deb_based
            install_deps_deb
    else
        die_distro
    fi
    install_venv
    configure_docker
    install_ovn_fake_multinode
    init_ovn_fake_multinode
    pull_ovn_fake_multinode
    install_ovn_tester
    pull_ovn_tester
    popd
}
```

We'll have to duplicate a bit of logic in the per-distro `install_deps_*` functions but, if we ensure that we test all supported distro types in CI, that might be acceptable.

What do you think?

File: ovn-fake-multinode-utils/playbooks/deploy-minimal.yml [hunk 9]
```diff
@@ -3,9 +3,8 @@
   become: true
   tasks:
   - name: Start openvswitch
-    systemd:
-      name: openvswitch
-      state: started
+    shell: |
+      systemctl start openvswitch || systemctl start openvswitch-switch
```
dceara, line 7: Nit: what if we use the `ansible_os_family` (https://docs.ansible.com/ansible/2.6/user_guide/playbooks_conditionals.html#ansible-os-family) conditional here and different tasks per OS family?

dceara: Using `ansible_os_family` would also allow us to use a a single `install-dependencies.yml` playbook.

ovn-fake-multinode-utils/playbooks/install-dependencies-deb.yml
```diff
@@ -0,0 +1,7 @@
+- name: Install dependencies
```
dceara, line 1: As mentioned above, I think we can merge this back into a single `install-dependencies.yml` where we add two conditional tasks (depending on OS family).  Wdyt?

ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
```diff
@@ -34,7 +33,7 @@
     shell: |
       sed -i '/^MaxSessions /d' /etc/ssh/sshd_config
       echo "MaxSessions 1024" >> /etc/ssh/sshd_config
-      systemctl reload sshd.service
+      systemctl reload sshd.service || systemctl reload ssh.service || true
```
dceara, line 36: Same nit as above about potentially using `ansible_os_family`.

do.sh
fnordahl: Using the `^ID` to get both the ID and ID_LIKE fields was intentional, that allows the check functions to iterate over all the ID's to help them decide what kind of distro we're dealing with. I'll add a comment to make that more clear.

do.sh
fnordahl: That does indeed look more appealing, makes it possible to glance at the `install` function and immediately understand what it does without all the implementation details. I'll redo this part, thx!

ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
fnordahl: I was wondering if there was a simple way of doing this with Ansible, and it looks like there is, I'll have a go at it. Thx for pointing it out!

do.sh
fnordahl: It turns out this *IS* required when using podman, otherwise we'll get the following while building the OVS/OVN containers which is done on the orchestrator:
```
Error: trying to reuse blob sha256:178c5a3203aea37107a1afcc9a965d6f9058cb12e31f6ecc900a48c310b27389 at destination: pinging container registry localhost:5000: Get "https://localhost:5000/v2/": http: server gave HTTP response to HTTPS client
```

I guess we can prune it to only hold the `[registries.insecure]` and `[registries.block]` sections?

ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
igsilya: Alternatively, maybe systemd module can correctly ignore re-starts of socket-activated services?
I.e. maybe we will not need to try to restart it twice and || true, if we split the task in two, one that updates teh config and one that restarts the daemon with systemd module.

I didn't try that though.

do.sh
dceara: Ok, but I'm not sure I understand where the iteration is.  Did you mean something like:
```
function is_rpm_based() {
    for i in $DISTRO_IDS; do
        case $i in
        centos* | rhel* | fedora*)
            true
            return
        ;;
        esac
    done
    false
}
```

fnordahl: So, I intuitively assumed the `in` operator would do both, and AFAICT it does?

```bash
#!/bin/bash

function is_abc() {
    echo "evaluatiing: '$DISTRO_IDS'"
    case $DISTRO_IDS in
        a* | b* | c*)
            echo a, b or c
        ;;
        *)
            echo neither
        ;;
    esac	
}

DISTRO_IDS="c"
is_abc
DISTRO_IDS="c b"
is_abc
DISTRO_IDS="c b
a"
is_abc
DISTRO_IDS="a b
c"
is_abc
DISTRO_IDS="d e
f"
is_abc
```

Output:
```
evaluatiing: 'c'
a, b or c
evaluatiing: 'c b'
a, b or c
evaluatiing: 'c b
a'
a, b or c
evaluatiing: 'a b
c'
a, b or c
evaluatiing: 'd e
f'
neither
```

fnordahl: Done

ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
fnordahl: Yup, I ended up doing that, and apparently the `state: reloaded` gracefully works fine with a socket activated service

ovn-fake-multinode-utils/playbooks/deploy-minimal.yml
fnordahl: Done

ovn-fake-multinode-utils/playbooks/install-dependencies-deb.yml
fnordahl: Done

do.sh
dceara: Thanks!

dceara (approved):
Looks good to me, thanks!

dceara:
@igsilya Do you want to have another look at these changes before I merge the PR?

igsilya:
> @igsilya Do you want to have another look at these changes before I merge the PR?

Sure.  I can take a look.  I won't test though. :)

ovn-fake-multinode-utils/playbooks/install-dependencies.yml
```diff
@@ -2,6 +2,24 @@
   hosts: ovn_hosts
   tasks:
     - name: Install required packages
-      shell: yum install git gcc openvswitch python3-pyyaml python3-virtualenv python3-devel containers-common --skip-broken -y
+      shell: |
+        yum install git gcc openvswitch python3-pyyaml python3-virtualenv \
+            python3-devel containers-common --skip-broken -y
```
igsilya, line 7: Can we use `ansible.builtin.package` module for this as well?
`ansible.builtin.dnf/yum` should also be fine, since we're in the OS-specific part here.

igsilya:
I added 2 small comments.  Otherwise LGTM.

ovn-fake-multinode-utils/playbooks/install-dependencies.yml
fnordahl: The reason for not doing that was the `--skip-broken` part in the original code.

If the reason for doing that is now gone, or we can accomplish something similar with Ansible, I'm fine with changing it. Anyone know why it was there in the first place?

igsilya: I'm not sure why it is there.  But `ansible.builtin.dnf` module has it: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html#parameter-skip_broken

fnordahl: Thx for the pointer!

After some light inspection, it does appear Ansible would potentially pass on task variables to whatever backend package module is used, ref: https://github.com/ansible/ansible/blob/bad8843124a50493141a3e3d7920353239021389/lib/ansible/plugins/action/package.py

I do not know for a fact that it works that way though. I'll try to add the `ansible.builtin.package` module and we'll see if that is inherited to the `dnf` module on RHEL systems?

I don't have one at hand so would appreciate if one of you would be able to check whether it works there?

ovn-fake-multinode-utils/playbooks/install-dependencies.yml
```diff
@@ -2,6 +2,24 @@
...
+      when: ansible_os_family == "RedHat"
+    - name: Install required packages
```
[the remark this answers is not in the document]
fnordahl: Done

ovn-fake-multinode-utils/playbooks/install-dependencies.yml
fnordahl: Done

ovn-fake-multinode-utils/playbooks/install-dependencies.yml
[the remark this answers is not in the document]
igsilya: I think, you missed the `name:` part.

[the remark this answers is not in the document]
igsilya: Ugh.  Wrong thread.

ovn-fake-multinode-utils/playbooks/install-dependencies.yml
igsilya: I think, you missed the `name:` part.

fnordahl: Nah, that didn't work, trying to duplicate the task with some some nitty gritty conditionals around RH distro specifics then.

igsilya:
```
Syntax Error while loading YAML.
  did not find expected '-' indicator

The error appears to be in '/tmp/cirrus-ci-build/ovn-fake-multinode-utils/playbooks/install-dependencies.yml': line 13, column 9, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

        - containers-common
        state: present
        ^ here
```

I it was the same error before.  I guess it should be:
```
 hosts: ovn_hosts
  tasks:
    - name: Install required packages
      ansible.builtin.package:
        name:
          - git
          - gcc
          - openvswitch
          - python3-pyyaml
          - python3-virtualenv
          - python3-devel
          - containers-common
        skip_broken: true
        state: present
      when: ansible_os_family == "RedHat"
```

fnordahl: On the [original change attempt](https://cirrus-ci.com/task/6206298660798464) it specifically choked on the `skip_broken` key

igsilya: Yes, because you didn't have a `name:` qualifier.  So, it tried to interpret `skip_broken` as part of the list.  If we have a `name:` and package list is indented, it shouldn't fail this way, AFAICT.

fnordahl: Ha, indeed, good spot, thx!

dceara: Unfortunately 'skip-broken' is broken (pun intended) on some ansible versions:
https://github.com/ansible/ansible/issues/73072

Officially this should've been starting with ansible 2.14.3 (https://github.com/ansible/ansible/commit/6bcb494f8306615f2b8741dad23529fdcd94626c) but I didn't manage to make it work on our RHEL8.7.  I'm using 

```
# ansible --version
ansible [core 2.14.3]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.13 (main, Nov  9 2022, 13:16:24) [GCC 8.5.0 20210514 (Red Hat 8.5.0-15)] (/usr/bin/python3.9)
  jinja version = 3.1.2
  libyaml = True
```

And still getting errors for missing packages:
```
TASK [Install required packages] *************************************************************************************************************************************************************************************************************
fatal: [advnetlab29.anl.eng.rdu2.redhat.com]: FAILED! => {"changed": false, "failures": ["No package openvswitch available."], "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}                            
fatal: [advnetlab11.anl.eng.rdu2.redhat.com]: FAILED! => {"changed": false, "failures": ["No package openvswitch available."], "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}
```

What if we use the following work around for now?

```
diff --git a/ovn-fake-multinode-utils/playbooks/install-dependencies.yml b/ovn-fake-multinode-utils/playbooks/install-dependencies.yml                                                                                                       
index 5229252..608e21b 100644
--- a/ovn-fake-multinode-utils/playbooks/install-dependencies.yml
+++ b/ovn-fake-multinode-utils/playbooks/install-dependencies.yml
@@ -6,15 +6,24 @@
         name:
           - git
           - gcc
-          - openvswitch
           - python3-pyyaml
           - python3-virtualenv
           - python3-devel
           - containers-common
-        skip_broken: true
         state: present
       when: ansible_os_family == "RedHat"

+    # 'skip-broken' might not work with specific ansible versions
+    # so ignore_errors for packages that might not be there.
+    # https://github.com/ansible/ansible/commit/6bcb494f8306
+    - name: Install optional packages
+      ansible.builtin.package:
+        name:
+          - openvswitch
+        state: present
+      ignore_errors: true
+      when: ansible_os_family == "RedHat"
+
     - name: Install required packages
       ansible.builtin.package:
         name:
```

fnordahl: Thx for testing @dceara, works for me, I'll include it in the PR

igsilya (approved):
Didn't test, but LGTM.

PR #159, opened 2023-03-10
Bring the base cluster up with ansible

igsilya:
This PR adds a playbook that will:

1.  Start an ovn-tester container on the first IP in mgmt_net.
2.  Start central nodes on IPs starting with the second IP.
3.  Start worker nodes equally distributed across all worker hosts.
4.  Start the process monitor script in each container, including the ovn-tester.
5.  Configure SSL remotes on ovsdb-server processes on workers.

This way ovn-tester no longer needs to call ovn-fake-multinode, set up SSL or start process monitoring.

Change allows to save a lot of time in high-scale tests since fake nodes are now deployed in parallel on each physical node.
For example, for us this saves 20+ minutes per 500-node test.


do.sh
```diff
@@ -389,8 +400,20 @@ function run_test() {
...
+    cluster_vars=""
+    for var in enable_ssl clustered_db monitor_all use_ovsdb_etcd \
+               node_net datapath_type n_relays n_workers; do
+        cluster_vars="${cluster_vars} $(get_cluster_var ${test_file} ${var})"
+    done
+    echo "-- Cluster vars: ${cluster_vars}"
+
+    if ! ansible-playbook ${ovn_fmn_playbooks}/bringup-cluster.yml \
+           -i ${hosts_file} --extra-vars "${cluster_vars}" ; then
+        die "-- Failed to bring up fake cluster!"
```
dceara, line 412: Should we try to collect some basic logs at this point?

dceara (commented):
I still need to test it out, so I just have one comment for now.  Looks great however!

do.sh
igsilya: Which ones?
FWIW, the output of the playbook will contain errors if there are some, and it will be dumped to the test-log.

dceara: Well, collect-logs.yml below should work just fine even if only part of the containers have been deployed.

igsilya: Well, I'm not convinced these logs contain anything interesting.  The cluster is not even connected yet.  Installation failures will mostly be in ansible failure logs.

Also, I'm not sure how to gracefully jump over to the log collection.  Do you have some idea how to do that?

dceara: You make a good point. I'm not sure how to nicely jump to the log collection part.  Let's forget about it for now and revisit it if we ever start having issues for which ansible failure logs are not enough.

igsilya:
Rebased on top of the current main.

igsilya:
@fnordahl , hi! Do you think it's OK if we require Ubuntu 22.10 as a minimum supported version?  There is an ansible version mess with all the collections re-naming.  See https://github.com/dceara/ovn-heater/pull/159/commits/f1822af3dfdeb38f8391eb0f47f37b814a77ef52.

fnordahl (approved):
LGTM, took it for a spin on a set of clean Ubuntu 22.10 machines and works as expected.

igsilya:
> LGTM, took it for a spin on a set of clean Ubuntu 22.10 machines and works as expected.

Thanks, @fnordahl !

I also tested on RHEL, and it works as expected there too.

dceara:
Thanks @igsilya and @fnordahl!

PR #162, opened 2023-05-15
do.sh: Improve host OS distro detection.

dceara:
Don't use `lsb_release` anymore and correctly detect Fedora/RHEL-like distro versions.

### DIFF
--- a/do.sh
+++ b/do.sh
@@ -40,6 +40,8 @@ USE_OVSDB_ETCD=${USE_OVSDB_ETCD:-no}
 # https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_05
 DISTRO_IDS=$(awk -F= '/^ID/{print$2}' /etc/os-release | tr -d '"')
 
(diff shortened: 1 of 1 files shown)


dceara:
@igsilya, wdyt?

do.sh
```diff
@@ -207,6 +231,7 @@ ovn_fmn_repo="${OVN_FAKE_MULTINODE_REPO:-https://github.com/ovn-org/ovn-fake-mul
 ovn_fmn_branch="${OVN_FAKE_MULTINODE_BRANCH:-main}"
 
 OS_IMAGE_OVERRIDE="${OS_IMAGE_OVERRIDE}"
+OS_IMAGE_DEFAULT="registry.fedoraproject.org/fedora:38"
```
igsilya, line 234: Before it was "whatever ovn-fake-multinode builds by default", which is `fedora:latest`.  Should we use `fedora:latest` here?

dceara: It makes sense.  I changed it back to `fedora:latest`.

igsilya (approved):
Didn't test, but looks OK.

PR #163, opened 2023-05-16
do.sh: Use built-in venv instead of virtualenv.

igsilya:
The functionality of built-in venv is enough for our use case.
Also, virtualenv is not available in some distributions (rhel9).


dceara (approved):
Looks good to me, thanks!

PR #164, opened 2023-05-17
ci: Fix logs checking in Cirrus-CI jobs.

dceara:
The check stage only cares about the last instruction's exit code. Run all checks inside a script that fails on the first error.


dceara:
CC: @igsilya

.cirrus.yml
```diff
@@ -57,11 +57,8 @@ low_scale_task:
         test-scenarios/ovn-low-scale.yml'
     - ./do.sh run test-scenarios/ovn-low-scale.yml low-scale
 
-  check_logs_script:
-    - '! grep -B 30 -A 30 "Result: FAIL" ./test_results/*/test-log'
-    - '! grep -B 30 -A 30 "Traceback" ./test_results/*/test-log'
-    - '! grep "Failed to run test" ./test_results/*/test-log'
-    - grep "Result: SUCCESS" ./test_results/*/test-log
+  check_foo_script:
```
igsilya, line 60: foo? :)

igsilya (changes requested):
The change didn't work, AFAICT: https://cirrus-ci.com/task/5962705555685376?logs=check_foo#L31

igsilya:
OK.  The main issue here is the negation.  It's treated as a subshell, so errexit is not effective.  We need to generate the exit manually it seems.

.cirrus.yml
dceara: Oops :)

dceara:
> OK. The main issue here is the negation. It's treated as a subshell, so errexit is not effective. We need to generate the exit manually it seems.

You're right.  I added explicit `exit`s now.

igsilya (approved):
I didn't test, but if ti works, then LGTM.

PR #166, opened 2023-05-26
Simplify deployment and some minor cleanups.

### LINKED ISSUE #130: Use a single container runtime and remove the need for a image registry.
We currently try to support multiple container runtimes (docker, podman, docker-ce) but that makes the code quite complex.  Maybe we should try to stick to a single runtime (e.g., podman).

We should also remove the need for a local image registry so we don't rely on "random" (e.g., from docker hub) registry images.  We can probably just ansible-copy the `ovn-multi-node` image do.sh builds to all DUT nodes instead of having them pull the image from a registry running on the tester node.

CC: @igsilya

dceara:
Closes #130


dceara:
Marking the PR as WIP as it seems the fedora CI is failing because of VMs running out of disk space.  Needs some more investigation.

dceara:
Marking the PR as WIP as it seems the fedora CI is failing because of VMs running out of disk space.  Needs some more investigation.

dceara:
It took a while to figure out the right magic to make CI happy.. It should be fine now, thanks @igsilya for pointing me at `growpart`.  It should be ready for review now.

.cirrus.yml
```diff
@@ -20,6 +20,23 @@ low_scale_task:
...
+  # /dev/<device><partition-number>[text]
+  resize_disk_script:
+    - d=$(findmnt -n -o SOURCE / | cut -f 1 -d '[' | cut -f 3 -d '/' | grep -oE '[a-zA-Z]+')
+    - p=$(findmnt -n -o SOURCE / | cut -f 1 -d '[' | cut -f 3 -d '/' | grep -oE '[0-9]+')
+    - t=$(df --output=fstype /root | grep -v Type)
+    - growpart /dev/$d $p || true
+    - '[ "$t" = "ext4" ] && resize2fs /dev/$d$p || true'
+    - '[ "$t" = "xfs" ] && xfs_growfs -d /root || true'
+    - '[ "$t" = "btrfs" ] && btrfs filesystem resize max /root || true'
+
```
igsilya, line 39: Shouldn't this be part of a previous commit that increased the disk size?

do.sh
```diff
@@ -322,11 +280,21 @@ function init_ovn_fake_multinode() {
...
 }
 
 function pull_ovn_tester() {
+    # Save the image locally, use docker-dir format because it allows
+    # compression.
+    rm -f ovn-tester-image.tar
+    podman save -o ovn-tester-image.tar ovn/ovn-tester:latest
+
     ansible-playbook ${ovn_fmn_playbooks}/pull-ovn-tester.yml -i ${hosts_file}
 }
```
igsilya, line 299: Could you add an echo saying what is happening before each  `podman save`?  It would be easier to understand what is going on.

Also the comment in `pull_ovn_tester()` is not correct as we're not using `docker-dir`.
Speaking of formats though, can we use oci-archive instead?  It provides compression for layers by default and is producing 3x smaller image:
```
-rw-r--r-- 1 root root 1.2G Jun  1 18:10 /tmp/docker-archive.tar
-rw-r--r-- 1 root root 468M Jun  1 18:09 /tmp/oci-archive.tar
```
Above are results for an actual `ovn-multi-node` image in different formats.  The size difference also explained here: https://github.com/containers/podman/discussions/17934
It takes a noticeable amount of time to sync 1.2 GB between many physical hosts.

.cirrus.yml
```diff
@@ -46,7 +64,7 @@ low_scale_task:
 
   pack_caches_script:
     - rm -rf runtime-cache/*
-    - docker save -o runtime-cache/ovn-multi-node.tar ovn/ovn-multi-node:latest
+    - podman save -o runtime-cache/ovn-multi-node.tar ovn/ovn-multi-node:latest
     - tar -czf runtime-cache/runtime.tar.gz runtime
```
igsilya, line 68: Why we're saving the image again here?  It should be already in `runtime/ovn-fake-multinode/ovn-multi-node-image.tar`.

igsilya (changes requested):
I added a couple of comments for image archives.  The rest looks good.  I also tested this in our lab and seems to work fine.

.cirrus.yml
dceara: You're right, it's there.  I had changed things so many times trying to figure out CI and in one of the versions I was (desperately) removing images from `runtime/..` to save disk space.  But that's not the case anymore.  I'll remove this and change the unpack_caches_script to load the image from the cached runtime.

do.sh
dceara: Cool, I didn't know about oci-archive.  I'll use that and fix all the other issues too, thanks!

.cirrus.yml
dceara: Done, seems to work:
https://cirrus-ci.com/task/5548665809403904?logs=unpack_caches#L2
https://cirrus-ci.com/task/6674565716246528?logs=unpack_caches#L2

do.sh
dceara: Done.

do.sh
```diff
@@ -320,13 +278,24 @@ function init_ovn_fake_multinode() {
...
+    pushd ${rundir}/ovn-fake-multinode
+    rm -f ovn-multi-node-image.tar
+    podman save --format oci-archive -o ovn-multi-node-image.tar \
+        ovn/ovn-multi-node:latest
     ansible-playbook ${ovn_fmn_playbooks}/pull-fake-multinode.yml -i ${hosts_file}
+    popd
 }
 
 function pull_ovn_tester() {
+    echo "-- Saving the ovn/ovn-tester image and pulling it on all nodes."
```
igsilya, line 294: s/all/tester/ ?

dceara: Ack. :)

igsilya (approved):
LGTM.  Thanks!

Issue #146, opened 2023-01-10
Tester crash on leadership change while adding LB to LS


igsilya:
```
2022-12-18 08:34:16,188 | ovn_workload |INFO| Binding lport lp-172-6 on ovn-scale-172
2022-12-18 08:34:16,310 | ovn_workload |INFO| Binding lport lp-172-7 on ovn-scale-172
2022-12-18 08:34:16,419 | ovn_workload |INFO| Binding lport lp-172-8 on ovn-scale-172
2022-12-18 08:34:16,564 | ovn_workload |INFO| Binding lport lp-172-9 on ovn-scale-172
2022-12-18 08:34:16,703 | ovsdbapp.backend.ovs_idl.vlog |INFO| ssl:192.16.0.2:6641: clustered database server is not cluster leader; trying another server
2022-12-18 08:34:16,703 | ovsdbapp.backend.ovs_idl.vlog |INFO| ssl:192.16.0.2:6641: connection closed by client
2022-12-18 08:34:16,706 | ovsdbapp.backend.ovs_idl.vlog |INFO| ssl:192.16.0.3:6641: connecting...
2022-12-18 08:34:16,710 | ovsdbapp.backend.ovs_idl.vlog |INFO| ssl:192.16.0.3:6641: connected
2022-12-18 08:34:16,717 | ovsdbapp.backend.ovs_idl.transaction |ERROR| Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/ovsdbapp/backend/ovs_idl/connection.py", line 118, in run
    txn.results.put(txn.do_commit())
  File "/usr/local/lib/python3.9/site-packages/ovsdbapp/backend/ovs_idl/transaction.py", line 92, in do_commit
    command.run_idl(txn)
  File "/usr/local/lib/python3.9/site-packages/ovsdbapp/schema/ovn_northbound/commands.py", line 1535, in run_idl
    raise RuntimeError("LB %s alseady exist in switch %s" % (
RuntimeError: LB 4835350f-8e4e-43c8-827f-97c405e25939 alseady exist in switch 6215d9e7-5d03-4b4e-b6ff-1bcc099fe35d

2022-12-18 08:34:16,717 | ovn_context  |INFO| Waiting for the OVN state synchronization
2022-12-18 08:34:17,091 | ovn_context  |INFO| Exiting context base_cluster_bringup
Traceback (most recent call last):
  File "/ovn-tester/ovn_tester.py", line 382, in <module>
[...]

PR #167, opened 2023-05-29
ovn-tester: Fix exceptions while adding LBs after leadership transfer.

igsilya:
Python IDL reconnects on leadership transfer before receiving reply for an in-progress transaction.  But the transaction might, and likely will, actually succeed.  So the LB may already exist after re-connection. Let ovsdbapp know about that, so it doesn't throw exceptions when this happens.

This doesn't solve all the possible issues with other types of resources, but LBs are the most common after ports that already have a proper handling in place.

Closes: https://github.com/ovn-org/ovn-heater/issues/146


dceara (approved):
Looks good to me, thanks!

PR #168, opened 2023-05-29
utils: Fix grep 3.8 warnings.

igsilya:
We only use egrep in one place, and it's not really needed there. egrep is deprecated and generates warnings:

`
  egrep: warning: egrep is obsolescent; using grep -E
`

Also the second `\` in `\-\-` may lead to undefined behavior and generates a warning now:

`
  grep: warning: stray \ before -
`


dceara (approved):
Looks good to me, thanks!

PR #171, opened 2023-08-11
ci: Bump Ubuntu to 23.03 and ensure pip is installed.

dceara:
22.10 is EOL and not usable in Cirrus CI anymore.  Also, ensure pip is installed in the venv on deb-based platforms too.

### DIFF
--- a/do.sh
+++ b/do.sh
@@ -108,16 +108,14 @@ function generate() {
 function install_deps_local_rpm() {
     echo "-- Installing local dependencies"
     yum install redhat-lsb-core datamash \
-        python3-pip python3-netaddr python3 python3-devel \
+        python3-netaddr python3 python3-devel \
         podman \
         --skip-broken -y
-    [ -e /usr/bin/pip ] || ln -sf /usr/bin/pip3 /usr/bin/pip
-
 }
 
 function install_deps_local_deb() {
     echo "-- Installing local dependencies"
-    apt -y install datamash podman python3-pip \
+    apt -y install datamash podman \
            python3-netaddr python3 python3-all-dev python3-venv
 }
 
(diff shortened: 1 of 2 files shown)


do.sh
```diff
@@ -134,9 +134,7 @@ function install_venv() {
         python3 -m venv ${ovn_heater_venv}
     fi
     source ${ovn_heater_venv}/bin/activate
-    if is_rpm_based; then
-        python3 -m ensurepip
-    fi
+    python3 -m ensurepip --upgrade
```
igsilya, line 137: Why this is needed at all?  Don't we install pip from the repo?

dceara: We do but apparently that's not enough to have it in the virtualenv too (at least after we bump to Ubuntu 23.04).  Without "ensurepip" I was getting:
https://cirrus-ci.com/task/5183715442163712?logs=install#L484
```
/root/ovn-heater/runtime/venv/bin/python3: No module named pip
```

igsilya: Hmm, should we not install it from the repo then?

dceara: We do:
https://github.com/ovn-org/ovn-heater/blob/32668b7bd4586db9cba16c02ea568a46fb272185/do.sh#L120
```
function install_deps_local_deb() {
    echo "-- Installing local dependencies"
    apt -y install datamash podman python3-pip \
           python3-netaddr python3 python3-all-dev python3-venv
}
```

That gets called before `install_venv()` where we do:
https://github.com/ovn-org/ovn-heater/blob/32668b7bd4586db9cba16c02ea568a46fb272185/do.sh#L136-L142
```
    source ${ovn_heater_venv}/bin/activate
    if is_rpm_based; then
        python3 -m ensurepip
    fi
    python3 -m pip install -r ${topdir}/utils/requirements.txt
    deactivate
    popd
```

Even in the error logs, we see pip was installed from the repo:
https://cirrus-ci.com/task/5183715442163712?logs=install#L399

For some reason it doesn't seem to appear installed in the venv.

igsilya: I saw that.  Let me re-phrase:
Should we remove python3-pip from an apt command line?

dceara: I can try, I'll re-push to see if CI agrees.

dceara: It passed just fine, I squashed the change in.  We now only install pip in the venv.

PR #174, opened 2023-09-14
Docs: Mention that TESTER node requires dedicated interface

mkalcok:
Similar to WORKER and CENTRAL nodes, TESTER also contacts OVN databases and therefore requires additional interface connected to the shared network


dceara (approved):
Downstream we normally run the tester bits on the same host as the ovn-central bits so we missed mentioning this in the documentation.  LGTM, thanks!

dceara:
@mkalcok  I also added you to the list of authors: https://github.com/ovn-org/ovn-heater/commit/417f10f248d7f89f769d95a5fc7bf87bb8263453

PR #175, opened 2023-09-14
[WIP/RFC] Extract generic methods from `WorkerNode` and `Cluster` classes.

fnordahl:
Note that this PR depends on and includes commits from #173


ovn-tester/ovn_workload.py
```diff
@@ -216,6 +196,123 @@ def wait(self, sbctl, timeout_s):
...
+    @ovn_stats.timeit
+    def provision_port(self, cluster, passive=False):
+        raise NotImplemented
+
+    @ovn_stats.timeit
+    def ping_external(self, cluster, port):
+        raise NotImplemented
+
+
+class WorkerNode(ChassisNode):
```
fnordahl, line 284: @dceara it is this inheritance that made me think of `ChassisNode` as `Compute` is not really fit with Kubernetes. Or what do you think?

fnordahl: It could also serve as a base for a `NetworkNode`

fnordahl:
To better understand how this could be used I've also created a OpenStack CMS plugin and test skeleton here: https://github.com/fnordahl/ovn-heater/commit/9e254e9471d52bb05d1cdf5706272305fbc514b0

ovn-tester/ovn_workload.py
dceara: I'm ok with both; if `ChassisNode` makes more sense to you let's go with that. :)

fnordahl:
The patches from this PR has been carried over to #179, let's review them there.

PR #178, opened 2023-09-21
do.sh: Add `refresh-tester` subcommand.

mkalcok:
`refresh-tester` subcommand of `do.sh` triggers only rebuild of a tester container. It provides a faster way to apply/test your changes if you work only inside the `ovn-tester` package, and you don't want to rebuild OVN packages or fake-multinode image.

---
This is just something that I found handy in my workflow.


dceara (approved):
Neat, thanks!

PR #179, opened 2023-09-22
Openstack cloud tests

mkalcok:
Please review commit by commit. This branch is based on #175 so the first 7 commits by @fnordahl can be skipped in review.

This change builds on separation of CMS implementations in #175 and it adds Openstack CMS. More info is in commit messages but at current stage it can simulate provisioning of:

- Project router
- Project's Internal network
- External network with gateway routing


ovn-tester/ovn_utils.py
```diff
@@ -431,28 +431,57 @@ def set_inactivity_probe(self, value):
             ("inactivity_probe", value),
         ).execute()
 
-    def lr_add(self, name):
+    def lr_add(
+        self, name, enabled: bool = False, ext_ids: Optional[Dict] = None
```
fnordahl, line 435: `enabled: bool = False`

This changes the behavior of the method and is the cause of the CI functional test failure.

Should this either be `enabled: Optional[bool] = None` or `enabled: bool = True`?

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,561 @@
...
+
+    def next_external_ip(self) -> DualStackIP:
+        """Return next available IP address from configured 'external_net'."""
+        self._ext_net_pool_index += 1
+        return self.cluster_cfg.external_net.forward(self._ext_net_pool_index)
+
+    def next_int_net(self):
+        """Return next subnet for project/tenant internal network.
+
+        This method takes subnet configured as 'node_net' in cluster config as
```
fnordahl, line 91: Note that I think many OpenStack clouds default to re-using the same CIDR for the internal network of every project.

Using different ones makes it simpler to have overview in a test though and allows for testing both NAT'ed and routed use cases.

I guess we need to consider whether having many networks using the same CIDR impacts scale and whether we need to test that somewhere down the line.

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,561 @@
...
+        for uuid_, port in network.ports.items():
+            if port.ip is None or port.ip == "unknown":
+                updated_port = self.nbctl.ls_port_set_ipv4_address(
+                    port, str(network.next_host_ip())
+                )
+                network.ports[uuid_] = updated_port
+                # XXX: Unable to update external_ids
+
+    def _add_network_port(self, network: LSwitch, project_id: str) -> LSPort:
+        """Create regular port in LSwitch with Neutron's external IDs."""
```
fnordahl, line 275: The settings for the port in this method looks like they come from the special Neutron OVN Metadata port.

For a regular VM port the type would be `""` and the `requested-chassis` would point at an actual node.

fnordahl (commented):
Thanks a lot for your work on this, it looks like a great start! I have left a couple of comments in-line and I'll continue to look later today

ovn-tester/cms/openstack/openstack.py
mkalcok: I think I just misunderstood the port that was created and thus I named it wrong. When user runs `openstack network create`, Neutron automatically creates one `LSPort` for the `LSwitch`. I wasn't entirely clear on the purposed of this port so I gave it this general name/description. Do you have any suggestion on naming/purpose of this port?

ovn-tester/ovn_utils.py
mkalcok: Thanks for the pointer, you are right. Without passing the `enabled` field, the router has "empty" value in the column which I mistook for the same thing as `enabled=false`

mkalcok: So I found a [source on this behavior](https://man7.org/linux/man-pages/man5/ovn-nb.5.html#Logical_Router_TABLE). empty/unset value in `enabled` column is equal to `true`.
Even though the Neutron sets the value to `true` explicitly, I'd say we can drop this change and let it be the default "unset".

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,561 @@
...
+
+        port = self.nbctl.ls_port_add(
+            lswitch=network,
+            name=str(uuid.uuid4()),
+            mac=str(RandMac()),
+            ext_ids=port_ext_ids,
+        )
+        self.nbctl.ls_port_set_set_type(port, "localport")
+        self.nbctl.ls_port_set_set_options(port, "requested-chassis=")
+        self.nbctl.ls_port_enable(port)
```
rjarry, line 297: This is a general remark. It looks like you are not using transactions, whereas the ML2 driver tries to do so as much as possible. I'm not sure if it has any performance implications. If not, maybe that's not important.

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,561 @@
...
+
+        return port
+
+    def _add_provider_network_port(self, network: LSwitch) -> LSPort:
+        """Add port to Logical Switch that represents connection to ext. net"""
+        options = (
+            "mcast_flood=false, mcast_flood_reports=true, "
+            "network_name=physnet1"
```
rjarry, line 305: At first, I though this was a typo, but it is handled as a special case. You pass it as a string and it is parsed as a `dict` internally. Weird...

rjarry (commented):
Hi,

thanks for working on this. I had gone another route (trying to mock some parts of neutron to reuse the actual ML2 driver code) which has proven unfruitful.

I don't see anything about the metadata agent here nor any southbound access to simulate what the ML2 driver does. Is there any plan to do so?

ovn-tester/cms/openstack/openstack.py
mkalcok: yeah, it's a weird inconsistency :D

ovn-tester/cms/openstack/openstack.py
fnordahl: Agreed, I think your approach is the best to start with as it is easier to work with and covers multiple use cases. If we come to the conclusion that creating multiple networks with the same CIDR is an important use case we can change it later.

ovn-tester/cms/openstack/openstack.py
mkalcok: Thanks for the tip about the transaction, I did not consider them. I assume that we would not be measuring/profiling the "cluster_bringup" but better performance never hurts. I'll look into where I could group calls into transactions :+1:

mkalcok:
Thanks for the reviews @fnordahl @rjarry.

> I don't see anything about the metadata agent here nor any southbound access to simulate what the ML2 driver does. Is there any plan to do so?

That's a good point. I focused on the NB because I assumed that that would be the primary database the Neutron would communicate with. I'll take a look at the SB as well. Do you have any pointers to relevant code/docs about this?

rjarry:
There are a two parts to consider:

1) the neutron server access to the southbound db: https://opendev.org/openstack/neutron/src/tag/21.1.2/neutron/scheduler/l3_ovn_scheduler.py

2) the metadata agent that runs on all compute nodes: https://opendev.org/openstack/neutron/src/tag/21.1.2/neutron/agent/ovn/metadata/agent.py

I am far from being an expert in this code but I know it has significant impact on scale because it multiplies the number of southbound db clients by two.

For performance reasons, openstack deployments usually recommend enabling ovn-controller `monitor_all=true` to reduce cpu pressure on the southbound ovsdb-server. The same logic is also applied for neutron metadata agent. To properly do scale testing, it would be good for ovn-heater to simulate these connections and subscriptions.

dceara:
> There are a two parts to consider:
> 
>     1. the neutron server access to the southbound db: https://opendev.org/openstack/neutron/src/tag/21.1.2/neutron/scheduler/l3_ovn_scheduler.py
> 
>     2. the metadata agent that runs on all compute nodes: https://opendev.org/openstack/neutron/src/tag/21.1.2/neutron/agent/ovn/metadata/agent.py
> 
> 
> I am far from being an expert in this code but I know it has significant impact on scale because it multiplies the number of southbound db clients by two (1 ovn-controller + 1 neutron metadata agent per compute node).
> 
> For performance reasons, openstack deployments usually recommend enabling ovn-controller `monitor_all=true` to reduce cpu pressure on the southbound ovsdb-server. The same logic is also applied for neutron metadata agent. To properly do scale testing, it would be good for ovn-heater to simulate these connections and subscriptions.

It seems to me we're only interested in how these extra read-only clients increase load on the Southbound.  If that's the case, we can simulate DB clients by running ovsdb-server relay instances.  Today we already start real SB relays if configured ([e.g., `cluster.n_relays > 0`](https://github.com/ovn-org/ovn-heater/blob/f19d060f8ea54650b8868272977aa345229d04a9/test-scenarios/ovn-low-scale.yml#L11)).  We can programmatically start more of these (without connecting ovn-controllers to them) and achieve the same type of load on the SB as with neutron agents.   IIRC @felixhuettner had also suggested this during one of the first ovn-heater community meetings.

felixhuettner:
> It seems to me we're only interested in how these extra read-only clients increase load on the Southbound. If that's the case, we can simulate DB clients by running ovsdb-server relay instances. Today we already start real SB relays if configured ([e.g., `cluster.n_relays > 0`](https://github.com/ovn-org/ovn-heater/blob/f19d060f8ea54650b8868272977aa345229d04a9/test-scenarios/ovn-low-scale.yml#L11)). We can programmatically start more of these (without connecting ovn-controllers to them) and achieve the same type of load on the SB as with neutron agents. IIRC @felixhuettner had also suggested this during one of the first ovn-heater community meetings.

I think that should give us a fairly accurate measurement of the neutron-ovn-metadata-agent. afaik it currently does not set monitor conditions, so we get a full db dump anyway (currently clarifying that).
The only thing we are loosing is the performance difference between a relay and the neutron agent when receiving updates from the ovsdb; but i don't think that should matter too much

ovn-tester/cms/openstack/tests/base_openstack.py
```diff
@@ -0,0 +1,28 @@
...
+        with Context(ovn, "base_cluster_bringup", worker_count) as ctx:
+            for i in ctx:
+                worker_node: ChassisNode = ovn.worker_nodes[i]
+                log.info(
+                    f"Provisioning {worker_node.__class__.__name__} "
+                    f"({i+1}/{worker_count})"
+                )
+                worker_node.provision(ovn)
+
+            _ = ovn.new_project(gw_nodes=ovn.worker_nodes)
```
fnordahl, line 28: I like this, it uses the same approach as the `base_clustrer_bringup` test that every ovn-kubernetes test case runs to initialize things before running other tests.

The `new_project` method should probably be called N times per test scenario where N is input from the test-scenario configuration without direct relation to the number of workers configured. Perhaps a `n_projects` key under `base_openstack`?

About the number of `gw_nodes`. Neutron currently has a hard coded maximum of 5 GW nodes per router, so we should probably mimic that behavior. The nodes are either selected by chance or by load (load as in number of routers hosted at any given priority)

felixhuettner:
> @dceara: I think the neutron metadata agent is not read-only, it also writes to the southbound db:
> 
> https://opendev.org/openstack/neutron/src/tag/21.1.2/neutron/agent/ovn/metadata/agent.py#L193-L196
> 
> https://opendev.org/openstack/neutron/src/tag/21.1.2/neutron/agent/ovn/metadata/agent.py#L292-L293

Yep,
there is also some regular heartbeat, but most people configure it 

> > It seems to me we're only interested in how these extra read-only clients increase load on the Southbound. If that's the case, we can simulate DB clients by running ovsdb-server relay instances. Today we already start real SB relays if configured ([e.g., `cluster.n_relays > 0`](https://github.com/ovn-org/ovn-heater/blob/f19d060f8ea54650b8868272977aa345229d04a9/test-scenarios/ovn-low-scale.yml#L11)). We can programmatically start more of these (without connecting ovn-controllers to them) and achieve the same type of load on the SB as with neutron agents. IIRC @felixhuettner had also suggested this during one of the first ovn-heater community meetings.
> 
> I think that should give us a fairly accurate measurement of the neutron-ovn-metadata-agent. it currently does not set monitor conditions, so we get a full db dump anyway. The only thing we are loosing is the performance difference between a relay and the neutron agent when receiving updates from the ovsdb; but i don't think that should matter too much

Allthough it seems like there is now some kind of monitor conditions going on here, but only for chassis https://opendev.org/openstack/neutron/src/tag/21.1.2/neutron/agent/ovn/metadata/ovsdb.py#L50-L53

ovn-tester/cms/openstack/tests/base_openstack.py
mkalcok: That's a good point about the GW nodes, thank you. I didn't know about the gw limit per router. It should be fairly easy to distribute chassis in balanced way

ovn-tester/cms/openstack/openstack.py
mkalcok: I've looked into this bit more today and I don't see a nice way to integrate transactions. Currently I'm using mainly `ovn_utils.OvnNbctl` moethods which are not "batch-able". I could get around it by using `OvnNbctl.idl` directly but then we'd lose code reusing.
Also, correct me if I'm wrong here, but ovn-heater does not measure/profile "cluster_bringup", does it? If not, I'd lean towards not implementing transactions now.

ovn-tester/ovn_utils.py
mkalcok: I've dropped the `enabled` parameter entirely, so I'll mark this conversation as resolved.

ovn-tester/cms/openstack/openstack.py
rjarry: I don't know if cluster bringup is measured by ovn-heater. But that's ok if we don't use transactions. At worst we'll get lower results. Since CMS's most likely use them, performance can only improve :)

ovn-tester/cms/openstack/tests/base_openstack.py
mkalcok: I had to eat my words about the easy redistribution mechanism :D balance distribution of individual GW priorities between bunch of routers turned out rather complex so I defaulted to using `random.sample()` :D

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,598 @@
...
+        This method adds new network to project with prefix 'ext_net',
+        adds it to the Project's router and configures default route on the
+        router to go through this network.
+
+        :param project: Project to which external network will be added
+        :param gw_nodes: List of chassis that act as network gateways.
+        :return: None
+        """
+
+        ext_net = self._create_project_net(f"ext_net_{project.uuid}", 1500)
```
felixhuettner, line 151: From my experience openstack environments have generally a low count of widely used external networks (sometimes just one). So i would here not generate a new external network, but rather randomly choose from a list of existing ones.

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,598 @@
...
+        ext_net_port = self._add_metadata_port(ext_net, project.uuid)
+        provider_port = self._add_provider_network_port(ext_net)
```
felixhuettner, line 153: as far as i know there will only ever be one provider network port per provider network. Since `_add_provider_network_port` sets this statically to `physnet1` it should only ever be called once

ovn-tester/cms/openstack/openstack.py
mkalcok: These are two very good points thank you. I'll decouple ext_net creation from project creation.

mkalcok:
@felixhuettner I decoupled external network creation from project creation and now the `BaseOpenstack` test creates only one ext_net and connects all projects to it.

I'm creating these suggested changes as separate commits for easier review. Please let me know if that's OK or if you'd like me to squash them in the end to [53339f9](https://github.com/ovn-org/ovn-heater/pull/179/commits/53339f9f79e8ba202a3e3d2fecbb54608243e08c)

felixhuettner:
> @felixhuettner I decoupled external network creation from project creation and now the `BaseOpenstack` test creates only one ext_net and connects all projects to it.
> 
> I'm creating these suggested changes as separate commits for easier review. Please let me know if that's OK or if you'd like me to squash them in the end to [53339f9](https://github.com/ovn-org/ovn-heater/pull/179/commits/53339f9f79e8ba202a3e3d2fecbb54608243e08c)

Thanks. Looks good

mkalcok:
I'd say that this change is now almost ready for full review. Two things that are missing and I'd like your opinion on are:

- Floating IPs and external connectivity: I'd be happy to do this in separate PR to not blow up scope of this one even more.
- ovn-metadata-agents: Is the consensus that we can simulate metadata agents with ovn-relay nodes to create "read" load on SB DB? Since as far as I can tell the only write operations of metadata agents are keepalive messages and updates to chassis external ID's on registration?

fnordahl:
> I'd say that this change is now almost ready for full review. Two things that are missing and I'd like your opinion on are:
> 
>     * Floating IPs and external connectivity: I'd be happy to do this in separate PR to not blow up scope of this one even more.
> 
>     * ovn-metadata-agents: Is the consensus that we can simulate metadata agents with ovn-relay nodes to create "read" load on SB DB? Since as far as I can tell the only write operations of metadata agents are keepalive messages and updates to chassis external ID's on registration?

From my perspective I think it would be valuable to finalize this PR as-is now and mark it as ready for review once you feel comfortable with it.

The items discussed above can be worked on independently, and the work you have done so far would be a dependency.

felixhuettner:
> @felixhuettner I decoupled external network creation from project creation and now the `BaseOpenstack` test creates only one ext_net and connects all projects to it.
> 
> I'm creating these suggested changes as separate commits for easier review. Please let me know if that's OK or if you'd like me to squash them in the end to [53339f9](https://github.com/ovn-org/ovn-heater/pull/179/commits/53339f9f79e8ba202a3e3d2fecbb54608243e08c)

I'm not sure what has happend, but we seem to create multiple external networks again (even though they now seem to have no external connectivity)?

mkalcok:
:facepalm: thank you for noticing. I lost a commit that decoupled project/ext net creation when I was rebasing/force pushing local copy on a different machine

felixhuettner (approved):

rjarry (approved):

ovn-tester/cms/openstack/tests/base_openstack.py
```diff
@@ -36,4 +37,4 @@ def run(self, ovn: OpenStackCloud, global_cfg):
                 worker_node.provision(ovn)
 
             for _ in range(self.config.n_projects):
-                _ = ovn.new_project(gw_nodes=ovn.worker_nodes)
+                _ = ovn.new_project(gw_nodes=self.config.n_gws_per_project)
```
fnordahl: I wonder if we should find a more specific name for this configuration option. While it fits well with the current implementation, it does not really fit with how it could be used in OVN.

The `Gateway_Chassis` (or possibly `HA_Chassis_Group` at some point in the future) is a property of the [`Logical_Router_Port`](https://manpages.ubuntu.com/manpages/jammy/man5/ovn-nb.5.html#logical_router_port%20table).

A project can have multiple routers, and routers can have multiple logical router ports with different configuration for these attributes. There is [ongoing work in OpenStack to support multiple LRPs per router](https://specs.openstack.org/openstack/neutron-specs/specs/2023.2/active-active-l3-gateway-with-multihoming.html).

Would perhaps `n_chassis_per_gw_lrp` work?

test-scenarios/openstack.yml
```diff
@@ -0,0 +1,11 @@
...
+  log_cmds: false
+  cms_name: openstack
+
+cluster:
+  clustered_db: true
+  log_txns_db: true
+  n_workers: 3
+
+base_openstack:
+  foo: bar
```
fnordahl: nit: newline at end of file please

fnordahl (commented):
Thank you for all your work on this, the proposal gives us a solid foundation (in record time) on which multiple people can iterate independently to implement the myriad of choices in OpenStack networking.

I have one comment about choice of configuration option name that I think we should address and a nit in-line.

We should probably also add this test case as a Cirrus CI job but I think that can be done as a follow-up.

Otherwise, this LGTM!

ovn-tester/cms/openstack/tests/base_openstack.py
mkalcok: hmm. If we call it `n_chassis_per_gw_rlp`, do we also need to tie it back to the projects with something like `n_gw_ports_per_project` ?

fnordahl: > hmm. If we call it `n_chassis_per_gw_rlp`, do we also need to tie it back to the projects with something like `n_gw_ports_per_project` ?

I'm very picky on configuration option naming, as opposed to implementations which change frequently, options are a kind of API and have a infinite lifetime.  Any future changes to their name or meaning may have expensive workflow change costs down the line.

So my intention with this comment was scoped at this specific option.

But you are right in that it would probably also make sense to have other options to control the adjacent entities in the project-router-lrp test matrix.

Would something like `n_gw_ports_per_router` and `n_routers_per_project` catch more use cases?

Not sure if we need them right now though.

mkalcok: You are right. We need to strike balance between future-proofing and not inventing unnecessary options here. We can stick with `n_chassis_per_gw_lrp` with implication that currently each project has only one router and one gw port.

mkalcok:
I pushed two small additions:

- Rename config option parameter according to @fnordahl suggestion
- Added task to `.cirrus.yml` to run low-scale Openstack test as part of the gate

felixhuettner (approved):

fnordahl:
I'm working on a rebase after #180 merged.

fnordahl:
Rebased and ready for review now @dceara @felixhuettner @rjarry

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,724 @@
...
+            "neutron:network_name": neutron_net.name,
+            "neutron:port_name": "",
+            "neutron:project_id": project_id,
+            "neutron:revision_number": "1",
+            "neutron:security_group_ids": neutron_net.security_group.name,
+            "neutron:subnet_pool_addr_scope4": "",
+            "neutron:subnet_pool_addr_scope6": "",
+        }
+        port_options = (
+            f"mcast_flood_reports=true "
```
booxter, line 446: This was removed lately https://review.opendev.org/c/openstack/neutron/+/888127

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,724 @@
...
+        }
+
+        port = self.nbctl.ls_port_add(
+            lswitch=network,
+            name=str(uuid.uuid4()),
+            mac=str(RandMac()),
+            ext_ids=port_ext_ids,
+        )
+        self.nbctl.ls_port_set_set_type(port, "localport")
+        self.nbctl.ls_port_set_set_options(port, "requested-chassis=")
```
booxter, line 488: I will remove the empty chassis key in ml2/ovn, it's not needed: https://review.opendev.org/c/openstack/neutron/+/897489

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,724 @@
...
+            "neutron:subnet_pool_addr_scope6": "",
+        }
+
+        port = self.nbctl.ls_port_add(
+            lswitch=network,
+            name=str(uuid.uuid4()),
+            mac=str(RandMac()),
+            ext_ids=port_ext_ids,
+        )
+        self.nbctl.ls_port_set_set_type(port, "localport")
```
booxter, line 487: (drive-by comment) these endpoints should probably be renamed to avoid double _set_.

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,724 @@
...
+        subnet_id = str(uuid.uuid4())
+        port_name = uuid.uuid4()
+        router_port_name = f"lrp-{port_name}"
+        device_owner = "network:router_gateway" if is_gw else ""
+        # XXX: neutron adds "neutron-" prefix to router name (which is UUID),
+        #      but for the ports' external IDs we want to extract just the
+        #      UUID.
+        router_id = router.name.lstrip("neutron-")
+
+        lsp_external_ids = {
```
booxter, line 579: (As a matter to discuss) A lot of effort here (passing parameters with "real" values, creating the dict) is spent to populate ext_ids with keys that are identical to a "real" openstack. But these keys are opaque for OVN, so we could shave off some complexity by either a) omitting populating the keys at all or b) if we want to make sure the db is populated with something, maybe just generate random strings / hardcode.

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,724 @@
...
+        snated_network = DualStackSubnet(int_net.cidr)
+        if snat_port is not None:
+            self.nbctl.nat_add(project.router, snat_port.ip, snated_network)
+
+    def add_vm_to_project(self, project: Project, vm_name: str):
+        compute = self.select_worker_for_port()
+        vm_port = self._add_vm_port(
+            project.int_net, project.uuid, compute, vm_name
+        )
+        compute.bind_port(vm_port)
```
booxter, line 304: we may want to also set mtu_request, as os-vif ovs port implementation does. This may affect behavior of OVN in some cases. (but this can be left to later.)

https://github.com/openstack/os-vif/blob/0a0dec37e42e220ffa750650027714bd7289faa1/vif_plug_ovs/ovsdb/ovsdb_lib.py#L68

ovn-tester/cms/openstack/openstack.py
fnordahl: We did ask ourselves whether it was worth populating them, and we concluded with three reasons to keep them:
1. To our surprise we found they are not entirely opaque, and there is actually processing of them in OVN [0][1].
2. Neutron's widespread use of external_ids increases the size of the database, we ought to measure what impact this has.
3. If you are developing/debugging the ovn-heater OpenStack plugin with a Neutron background, having these familiar keys with data that actually link to something can be useful.

0: https://github.com/ovn-org/ovn/blob/430c0fd32dd709b8d452ce9a339dd30228744b62/northd/northd.c#L1114-L1120
1: https://github.com/ovn-org/ovn/blob/430c0fd32dd709b8d452ce9a339dd30228744b62/northd/northd.c#L3683-L3690

ovn-tester/ovn_workload.py
```diff
@@ -863,38 +388,25 @@ def ping_ports(self, ports):
...
-        self.load_balancer.add_vips(vips)
-
-    def unprovision_vips(self):
-        if self.load_balancer:
-            self.load_balancer.clear_vips()
-            self.load_balancer.add_vips(self.cluster_cfg.static_vips)
-        if self.load_balancer6:
-            self.load_balancer6.clear_vips()
-            self.load_balancer6.add_vips(self.cluster_cfg.static_vips6)
+    def mesh_ping_ports(self, ports: List[ovn_utils.LSPort]) -> None:
```
dceara, line 391: Is it worth accounting(`@ovn_stats.timeit`) how long the mesh_ping took in total?

ovn-tester/cms/openstack/openstack.py
fnordahl: I agree that looks weird, I would like to propose we address that pre-existing issue in `ovn-tester/ovn_utils.py` separately though.

fnordahl:
@booxter @dceara for when you have a spare moment, I believe all review feedback has been addressed.

felixhuettner (approved):

booxter (approved):
Code wise I don't have any particular issues. Also, I believe we can iterate on it later and it's better to merge this PR that is already quite huge.

(I assume it doesn't break existing k8s scenarios.)

dceara:
@mkalcok @fnordahl There are a few conflicts due to merging the initial ovn-ic work; I'll try to help with the rebase.

dceara:
I tried a rebase here (completely untested yet): https://github.com/dceara/ovn-heater-1/commits/openstack-cloud-setup-rebase-tmp

fnordahl:
> I tried a rebase here (~completely untested yet~ barely tested): https://github.com/dceara/ovn-heater-1/commits/openstack-cloud-setup-rebase-tmp

Thanks, I'll have a look

fnordahl:
@dceara thanks a lot for the help with the rebase. I put your proposals for adaption into the correct commits and attributed accordingly.

I redid the parts of the commits that mainly move code around to ensure no changes were lost.

I also arrived at a slightly different solution to calling the `prepare_test` method, hope you agree.

Looking forward to your review!

fnordahl:
I just noticed that the rebase is one commit shorter than the previous iteration, and the reason is that I have unintentionally squashed "CMS Openstack: Add ability to simulate VMs" and "OS Test: Spawn VMs and run ping in base scenario".

However, the result does look fine to me as both commits kind of addressed the same thing in iterations, so perhaps we should just keep it this way? Thoughts @mkalcok ?

mkalcok:
yeah, I'm fine with them being squashed. I separated them for easier review as one added logic to the `CMS Openstack` implementation and the other used that implementation in a test scenario, but they are closely related.

ovn-tester/ovn_utils.py
```diff
@@ -5,7 +5,7 @@
 import time
```
dceara, line 5: Thanks a lot for adding the type hints!

ovn-tester/cms/openstack/openstack.py
```diff
@@ -0,0 +1,555 @@
...
+        # Allocate worker IPs after central and relay IPs.
+        mgmt_ip = (
+            self.cluster_cfg.node_net.ip
+            + 2
+            + len(self.central_nodes)
+            + len(self.relay_nodes)
+        )
+
+        protocol = "ssl" if self.cluster_cfg.enable_ssl else "tcp"
+        # XXX: To facilitate network nodes we'd have to make bigger change
```
dceara, line 121: We had some discussions internally about potentially removing the ovn-fake-multinode dependency and start OVN components directly from ovn-heater.  Whenever (if) we do that it might be a good moment for adding support for network nodes too.

ovn-tester/cms/openstack/openstack.py
```diff
@@ -234,6 +241,38 @@ def add_internal_network_to_project(
         if snat_port is not None:
             self.nbctl.nat_add(project.router, snat_port.ip, snated_network)
 
+    def _get_gateway_chassis(self, count: int = 1) -> List[ChassisNode]:
+        """Return list of Gateway Chassis with size defined by 'count'.
+
+        Chassis are picked at random from all available 'worker_nodes' to
```
dceara, line 316: Nit: I understand the intent but on the other hand, this is a testing tool and reproducible runs are very important .  Not necessarily in this PR but do we need to log the chosen values and provide a way to explicitly select the same chassis if we want to re-run the exact same workload while debugging a DUT issue?

ovn-tester/ovn_utils.py
```diff
@@ -438,28 +438,53 @@ def set_inactivity_probe(self, value):
...
+        uuid = self.uuid_transaction(
+            partial(self.idl.lr_add, name, external_ids=ext_ids)
+        )
         return LRouter(name=name, uuid=uuid)
 
-    def lr_port_add(self, router, name, mac, dual_ip=None):
+    def lr_port_add(
+        self, router, name, mac, dual_ip=None, ext_ids: Optional[Dict] = None
+    ):
+        ext_ids = {} if ext_ids is None else ext_ids
```
dceara, line 472: We're missing:
```
options = {} if options is None else options
```

which causes:
https://cirrus-ci.com/task/5461921055375360?logs=test_openstack#L447
```
2023-10-25 11:40:24,426 | ovsdbapp.backend.ovs_idl.vlog |ERROR| attempting to write bad value to column options (ovsdb error: expected string, got <class 'NoneType'>)
```

@mkalcok I can squash it in if you agree with this change.

dceara (approved):
This is awesome work!  Thanks a lot @mkalcok @fnordahl @booxter @rjarry @felixhuettner for all the code and reviews!

I agree with @booxter let's get this in and we can address leftovers/known issues/bugs with follow up PRs.

There is one thing I'd like to fix first before merging this but I'll wait for a reply from @mkalcok or @fnordahl on https://github.com/ovn-org/ovn-heater/pull/179#pullrequestreview-1700628437

Thanks,
Dumitru

ovn-tester/cms/openstack/openstack.py
mkalcok: This is a very good point. I originally intended to do even distribution of GWs but the implementation turned out more complicated than I anticipated so I went with random. I'll open a ticket about this so that the issue is not lost.

ovn-tester/ovn_utils.py
mkalcok: Yes, thank you for noticing this. I seem to remember fixing this one already, not sure how it resurfaced. I'll take care of it.

fnordahl: > Yes, thank you for noticing this. I seem to remember fixing this one already, not sure how it resurfaced. I'll take care of it.

No wonder you thought that, because apparently I regressed this in "cms/openstack: Set MTU for VM and GW ports."! Sorry about that and thank you for fixing.

dceara: Thanks, looks good now!

ovn-tester/cms/openstack/openstack.py
dceara: Thanks @mkalcok

PR #180, opened 2023-09-28
Refactor cluster bringup.

dceara:
Define explicit nodes for all components (central DBs, relays, etc.), avoid IP guesswork and fix a few bugs.

### DIFF
--- a/ovn-tester/ovn_workload.py
+++ b/ovn-tester/ovn_workload.py
@@ -52,15 +52,16 @@ BrExConfig = namedtuple('BrExConfig', ['physical_net'])
 
 
 class Node(ovn_sandbox.Sandbox):
-    def __init__(self, phys_node, container, mgmt_ip):
+    def __init__(self, phys_node, container, mgmt_ip, protocol):
         super().__init__(phys_node, container)
         self.container = container
         self.mgmt_ip = netaddr.IPAddress(mgmt_ip)
+        self.protocol = protocol
 
 
 class CentralNode(Node):
-    def __init__(self, phys_node, container, mgmt_ip):
-        super().__init__(phys_node, container, mgmt_ip)
+    def __init__(self, phys_node, container, mgmt_ip, protocol):
+        super().__init__(phys_node, container, mgmt_ip, protocol)
 
     def start(self, cluster_cfg, update_election_timeout=False):
         log.info('Configuring central node')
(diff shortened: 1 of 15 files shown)


dceara:
CC: @LorenzoBianconi - this probably should go in before your PR #169 
CC: @fnordahl @mkalcok - this probably will conflict with your PR #179 but I think it's better to tackle these bringup issues early, before we add a new CMS.

mkalcok:
Thanks for the heads-up @dceara . I'll rebase/fix #179 once this is merged.

dceara:
> Thanks for the heads-up @dceara . I'll rebase/fix #179 once this is merged.

Thanks, if you have some time to review #180 that would also be awesome. :)

ovn-tester/ovn_workload.py
```diff
@@ -147,6 +139,27 @@ def central_containers(self):
         return self.db_containers
 
 
+class RelayNode(Node):
+    def __init__(self, phys_node, container, mgmt_ip):
```
mkalcok: This is slightly unrelated, but I was thinking whether there's a good way to improve semantics/logic of the `container` argument in the `Node` subclasses. All three `WorkerNode`, `CentralNode` and `RelayNode` accept `container` argument but it doesn't really affect anything. When creating these classes you have to match what's pretty much hardcoded. For the [relay](https://github.com/ovn-org/ovn-fake-multinode/blob/af8f4579bc487ad56bd7d7e024d627d9ea4a32a5/ovn_cluster.sh#L49) and [central](https://github.com/ovn-org/ovn-fake-multinode/blob/af8f4579bc487ad56bd7d7e024d627d9ea4a32a5/ovn_cluster.sh#L21), prefix for the container names is set in ovn-fake-multinode script and for the worker nodes it can be set via [prefix](https://github.com/ovn-org/ovn-heater/blob/6ad787025dd8027bb07185c4cca44c20e77ecb58/ovn-fake-multinode-utils/generate-hosts.py#L65) config option.

I think it'd be better if the logic of creating instances of these classes was entirely hidden from user and they'd be created automatically based on the values from test scenario and configs. Right now these classes give illusion of control but in reality user has to match names and counts defined somewhere else.

This is of course topic for entirely separate change, just something I was thinking about.

mkalcok (approved):
Although I'm not expert on ovn-heater, I'd say that overall LGTM, nice changes.

ovn-tester/ovn_workload.py
```diff
@@ -56,7 +56,7 @@ class Node(ovn_sandbox.Sandbox):
     def __init__(self, phys_node, container, mgmt_ip):
         super().__init__(phys_node, container)
         self.container = container
-        self.mgmt_ip = mgmt_ip
+        self.mgmt_ip = netaddr.IPAddress(mgmt_ip)
```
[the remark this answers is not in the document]
dceara: Yes, that's exactly it.  I was doing something like that while trying out @LorenzoBianconi's ovn-ic changes, using a netaddr.IPAddress externally through iterate through node IPs and ending up changing all the ones that were already assigned.

ovn-tester/ovn_workload.py
dceara: That's a very good point.  I'll see if it's easy to handle with a new commit in this PR, otherwise I'll open an issue to follow up later, thanks!

PR #182, opened 2023-10-19
bringup-cluster: Compute central facts for all types of nodes.

dceara:
We accidentally only computed them for central_hosts but we actually need them on workers too.

Fixes: 29e078134421 ("ovn-tester: add the capability to run multiple independent clusters")


LorenzoBianconi:
@dceara it seems fine to me

LorenzoBianconi (commented):
@dceara it seems correct to me

PR #183, opened 2023-10-22
Fix IC address set usage.

dceara:

### DIFF
--- a/ovn-tester/cms/ovn_kubernetes/ovn_kubernetes.py
+++ b/ovn-tester/cms/ovn_kubernetes/ovn_kubernetes.py
@@ -70,13 +70,13 @@ class Namespace:
         # Always add port IPs to the address set but not to the PGs.
         # Simulate what OpenShift does, which is: create the port groups
         # when the first network policy is applied.
-        if self.addr_set4:
-            for i, nbctl in enumerate(self.nbctl):
+        for i, nbctl in enumerate(self.nbctl):
+            if self.addr_set4[i]:
                 nbctl.address_set_add_addrs(
                     self.addr_set4[i], [str(p.ip) for p in ports]
                 )
-        if self.addr_set6:
-            for i, nbctl in enumerate(self.nbctl):
+        for i, nbctl in enumerate(self.nbctl):
+            if self.addr_set6[i]:
                 nbctl.address_set_add_addrs(
                     self.addr_set6[i], [str(p.ip6) for p in ports]
                 )
@@ -98,9 +98,9 @@ class Namespace:
             nbctl.port_group_del(self.pg_def_deny_igr[i])
             nbctl.port_group_del(self.pg_def_deny_egr[i])
             nbctl.port_group_del(self.pg[i])
-            if self.addr_set4:
+            if self.addr_set4[i]:
                 nbctl.address_set_del(self.addr_set4[i])
-            if self.addr_set6:
+            if self.addr_set6[i]:
                 nbctl.address_set_del(self.addr_set6[i])
             nbctl.port_group_del(self.sub_pg[i])
             nbctl.address_set_del(self.sub_as[i])


dceara:
CC: @LorenzoBianconi @igsilya

ovn-tester/ovn_workload.py
```diff
@@ -494,13 +494,13 @@ def add_ports(self, ports, az=0):
         # Always add port IPs to the address set but not to the PGs.
         # Simulate what OpenShift does, which is: create the port groups
         # when the first network policy is applied.
-        if self.addr_set4:
-            for i, nbctl in enumerate(self.nbctl):
+        for i, nbctl in enumerate(self.nbctl):
```
LorenzoBianconi, line 497: @dceara re-looking at the code, should we add port just to the address add_set{4,6}[az]?

ovn-tester/ovn_workload.py
```diff
@@ -522,9 +522,9 @@ def unprovision(self):
             nbctl.port_group_del(self.pg_def_deny_igr[i])
             nbctl.port_group_del(self.pg_def_deny_egr[i])
             nbctl.port_group_del(self.pg[i])
-            if self.addr_set4:
+            if self.addr_set4[i]:
```
LorenzoBianconi, line 525: @dceara ack, fine.

.cirrus.yml
```diff
@@ -72,8 +72,22 @@ low_scale_task:
   test_script:
     - 'sed -i "s/^  log_cmds\: False/  log_cmds\: True/"
         test-scenarios/ovn-low-scale*.yml'
-    - ./do.sh run test-scenarios/ovn-low-scale.yml low-scale
-    - ./do.sh run test-scenarios/ovn-low-scale-ic.yml low-scale-ic
+    - 'for ip4 in True False; do
+        for ip6 in True False; do
```
igsilya, line 76: Can we create a matrix instead of putting 3 different tests into one?

ovn-tester/ovn_workload.py
dceara: To simulate what ovn-k8s does we should add the IPs to address sets in all AZs.

LorenzoBianconi: @dceara ack, fine in this case

.cirrus.yml
dceara: It took some effort and it has some downsides but I pushed it to the PR as it is cleaner.  One thing we might have to adjust is the matrix size.  Especially with the thought in mind that we'll be adding OpenStack scenarios too.

.cirrus.yml
```diff
@@ -53,6 +39,26 @@ low_scale_task:
...
+      - image_project: ubuntu-os-cloud
+        image: family/ubuntu-2304-amd64
+    platform: linux
+    memory: 8G
+    disk: 40
+
+  runtime_cache:
+    folder: runtime-cache
+    fingerprint_script:
+      - echo ${CIRRUS_BUILD_ID}
```
igsilya, line 58: This is a little concerning.   We could consider going back to a simple loop, but we could also consider having a separate global and local to the job caches to save CPU cycles.

.cirrus.yml
```diff
@@ -69,11 +75,65 @@ low_scale_task:
...
+
+  compute_engine_instance:
+    matrix:
+      - image_project: fedora-cloud
+        image: family/fedora-cloud-38
+      - image_project: ubuntu-os-cloud
+        image: family/ubuntu-2304-amd64
+    platform: linux
+    memory: 8G
+    disk: 40
```
igsilya, line 92: Looking at the resource usage, we could, likely reduce the cpu count from the default 2 to 1.
We probably don't need 8G of RAM as well, but I'm not sure why ubuntu builds eat almost 4, while fedora builds only use ~500 MB.

.cirrus.yml
dceara: I'll remove the commit that increases CI coverage from this PR.  We need the address set fix sooner rather than later.  I'll open an issue to track adding global and local to the job caches to save CPU cycles.

PR #189, opened 2023-11-10
Add type hints for parameters and return

linup2011:
I've run following tests with the patch:
./do.sh run test-scenarios/ovn-20.yml ovn-20
./do.sh run test-scenarios/ovn-low-scale.yml ovn-low-scale
./do.sh run test-scenarios/ovn-low-scale-ic.yml ovn-low-scale-ic


numansiddique:
This PR made me learn what type hints are.

I've no objections to it.  If this PR is accepted,  I'd suggest make it a norm (or must) for new PRs to follow type hinting.
If it is too much of an ask for other contributors then perhaps its better to drop this PR.

Thanks

dceara:
> This PR made me learn what type hints are.
> 
> I've no objections to it. If this PR is accepted, I'd suggest make it a norm (or must) for new PRs to follow type hinting. If it is too much of an ask for other contributors then perhaps its better to drop this PR.
> 
> Thanks

Good point, there actually are ways to enforce this.  One way is to run `mypy .. ` as a GitHub action check on new PRs.

ovn-tester/ovn_tester.py
```diff
@@ -36,7 +37,9 @@ def usage(name):
     )
 
 
-def read_physical_deployment(deployment, global_cfg):
+def read_physical_deployment(
+    deployment: str, global_cfg: GlobalCfg
+) -> Tuple[PhysicalNode, List[PhysicalNode]]:
```
dceara, line 42: This actually returns `Tuple[List[PhysicalNode], List[PhysicalNode]]`.

linup2011: done

almusil:
> > This PR made me learn what type hints are.
> > I've no objections to it. If this PR is accepted, I'd suggest make it a norm (or must) for new PRs to follow type hinting. If it is too much of an ask for other contributors then perhaps its better to drop this PR.
> > Thanks
> 
> Good point, there actually are ways to enforce this. One way is to run `mypy .. ` as a GitHub action check on new PRs.

It makes sense, however if we want to check those rules it should be part of the PR so there is no chance to "break" in the meantime.

dceara:
> > > This PR made me learn what type hints are.
> > > I've no objections to it. If this PR is accepted, I'd suggest make it a norm (or must) for new PRs to follow type hinting. If it is too much of an ask for other contributors then perhaps its better to drop this PR.
> > > Thanks
> > 
> > 
> > Good point, there actually are ways to enforce this. One way is to run `mypy .. ` as a GitHub action check on new PRs.
> 
> It makes sense, however if we want to check those rules it should be part of the PR so there is no chance to "break" in the meantime.

Sure, that's the ideal way.  However, I think it's a pity to delay this PR for that.  I opened an issue to track running and enforcing rules with mypy on new PRs: https://github.com/ovn-org/ovn-heater/issues/192

dceara (approved):

PR #191, opened 2024-01-03
playbooks/dependencies: Update package cache on deb systems.

fnordahl:
If the package cache is not updated for example after a clean install, the `install-dependencies` playbook may fail with errors like this:

    TASK [Install container command] ***************************
    {"changed": false, "msg": "No package matching 'podman' is available"}

Use the `apt` build in ansible module and set `update_cache` to 'true' to fix this issue.


dceara (approved):
LGTM, thanks!

PR #194, opened 2024-03-04
do.sh: Fix bloating of the resource usage report for central nodes.

igsilya:
In case of a large number of availability zones the size of the resource usage report becomes unmanageable and browsers are not even able to open them.  For example, the size of the report on a 120 node cluster-density test is 166 MB.

Fix that the same way we do for workers - gather statistics only from the first 3 zones.

For the future we'll need to have aggregated stats in addition to these truncated ones, so we don't overlook potential issues and can compare cluster-wide resource usage with different number of zones.

Fixes: 40716b0bef06 ("ovn-tester: add the capability to deploy central components on multiple hv for ovn-ic")


dceara (approved):
LGTM, thanks!

PR #196, opened 2024-03-14
density_heavy: density_light: Fix startup stages when running in IC mode.

dceara:
In this mode we should actually distribute the ports that are created in the startup stage between AZs (clusters) instead of creating all of them in all AZs.

Fixes: 2562afb45721 ("ovn-tester: extend density_light testing for ovn-ic")
Fixes: 84649236d689 ("ovn-tester: extend density_heavy for ovn-ic")

### DIFF
--- a/ovn-tester/ovn_utils.py
+++ b/ovn-tester/ovn_utils.py
@@ -912,3 +912,8 @@ class OvnIcNbctl:
     def ts_add(self):
         log.info('Creating transit switch')
         self.uuid_transaction(partial(self.idl.ts_add, 'ts'))
+
+
+def distribute_n_tasks_per_clusters(n_tasks, n_clusters):
+    div, rest = divmod(n_tasks, n_clusters)
+    return [div + 1 if i < rest else div for i in range(n_clusters)]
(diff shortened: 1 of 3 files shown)


ovn-tester/ovn_utils.py
```diff
@@ -912,3 +912,8 @@ def uuid_transaction(self, func):
     def ts_add(self):
         log.info('Creating transit switch')
         self.uuid_transaction(partial(self.idl.ts_add, 'ts'))
+
+
+def distribute_n_tasks_per_clusters(n_tasks, n_clusters):
+    div, rest = divmod(n_tasks, n_clusters)
+    return [div + 1 if i < rest else 0 for i in range(n_clusters)]
```
LorenzoBianconi, line 919: @dceara I think this is not correct, e.g: 
for div = 2, rest = 3, it will end up in [3, 3, 3, 0, 0, 0].

I think it should be something like:
return [div + 1 if i < rest else div for i in range(n_clusters)]

dceara: Good catch, the precedence of operations in Python got me.  I'll fix this.

LorenzoBianconi (approved):
Thx for fixing it

Issue #195, opened 2024-03-04
Add aggregate max/avg/sum graphs to resource usage reports.


igsilya:
We're currently collecting stats only from the first 3 fake nodes.  While all of them should be equal, we can still overlook some potential issues where it is not the case.  We can't just collect the data from all the nodes, the reports will become unreadbale and too large for browsers to open, but we could try to plot max, avg and sum for all the components by their type.

The main challenges would be to try and map data points taken from not exactly the same point in time into a single aggregate point, potentially by interpolating the data points filling the gaps in the grid.  And another tricky part is to distinguish between different process types, e.g. we have multiple ovsdb-server processes in each central fake node and their pids are not always aligned.

igsilya:
I'll try to work on this.

PR #198, opened 2024-03-27
Add aggregate resource usage graphs

igsilya:
This PR adds a new html report `resource-usage-report-aggregate.html` that contains graphs for aggregate memory and CPU usage per process type as well as numbers for peak memory and CPU usage of all OVN processes combined.

`pandas.DataFrame` does all the heavy lifting of data interpolation between uneven timestamps and aggregation across processes.

Closes: #195

### DIFF
--- a/ovn-fake-multinode-utils/process-monitor.py
+++ b/ovn-fake-multinode-utils/process-monitor.py
@@ -36,7 +36,14 @@ def monitor(suffix: str, out_file: str, exit_file: str) -> None:
             tme = time.time()
             for p in processes:
                 try:
-                    name = p.name() + "-" + suffix + "-" + str(p.pid)
+                    name = p.name()
+                    for arg in p.cmdline():
+                        if arg.endswith('.pid') or arg.endswith('.py'):
+                            name = arg.split('/')[-1].split('.')[0]
+                            break
+
+                    name = name + "|" + suffix + "|" + str(p.pid)
+
                     # cpu_percent(seconds) call will block
                     # for the amount of seconds specified.
                     cpu = p.cpu_percent(0.5)
--- a/ovn-fake-multinode-utils/process-monitor.py
+++ b/ovn-fake-multinode-utils/process-monitor.py
@@ -22,7 +22,7 @@ def monitor(suffix: str, out_file: str, exit_file: str) -> None:
             for p in psutil.process_iter():
                 if any(name in p.name() for name in process_names):
                     processes.add(p)
-                elif any(
+                elif p.name() != 'monitor' and any(
                     name in part
                     for part in p.cmdline()
                     for name in process_names
--- a/utils/process-stats.py
+++ b/utils/process-stats.py
@@ -1,11 +1,17 @@
+import argparse
 import json
+import logging
 import os
 import pandas as pd
 import plotly.express as px
 import sys
 
-from datetime import datetime
 from typing import Dict, List
+
+
+FORMAT = '%(asctime)s |%(levelname)s| %(message)s'
+logging.basicConfig(stream=sys.stdout, level=logging.INFO, format=FORMAT)
+log = logging.getLogger(__name__)
 
 
 def read_file(filename: str) -> Dict:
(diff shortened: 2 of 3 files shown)


utils/process-stats.py
```diff
@@ -13,7 +13,41 @@ def read_file(filename: str) -> Dict:
...
-def resource_stats_generate(filename: str, data: Dict) -> None:
+def aggregated(df: pd.DataFrame) -> (pd.DataFrame, int):
+    column_names = list(df.columns)
+
+    df = df.pivot(
+        index='Time', columns='Process', values=column_names[2]
+    ).interpolate(method='time', limit_direction='both')
+
+    result = pd.DataFrame(index=df.index)
+    processes = set([p.split('|')[0] for p in list(df.columns)])
```
dceara, line 24: No need for list comprehension.  I think it can just be:
```
processes = {p.split('|')[0] for p in list(df.columns)}
```

utils/process-stats.py
```diff
@@ -27,19 +61,48 @@ def resource_stats_generate(filename: str, data: Dict) -> None:
...
+        x='Time',
+        y='CPU (%)',
+        color='Process',
+        title=('Aggregate ' if aggregate else '') + 'CPU usage',
     )
 
     with open(filename, 'w') as report_file:
         report_file.write('<html>')
+        if aggregate:
+            report_file.write(
```
dceara, line 86: Eww, but OK. :)

igsilya: I replaced this with f-string, but it's still not very pretty. :)

dceara (approved):
Thanks @igsilya!  This looks good to me!

utils/process-stats.py
```diff
@@ -60,17 +133,36 @@ def resource_stats_generate(filename: str, data: Dict) -> None:
...
-        sys.exit(1)
+    parser = argparse.ArgumentParser(
+        description='Generate resource usage charts.'
+    )
+    parser.add_argument(
+        '--aggregate', action='store_true', help='generate aggregate charts'
+    )
+    parser.add_argument(
+        '-o', '--output', required=True, help='file to write an HTML result'
+    )
```
putnopvut, line 144: Since this argument is required, it seems odd to have the `-o` prefix for the option. Perhaps instead this could just be
```
    parser.add_argument(
        'output', required=True, help='file to write an HTML result'
    )
```

This would also allow for existing callers of `process_stats.py` to not have to be updated.

ovn-fake-multinode-utils/process-monitor.py
```diff
@@ -36,7 +36,14 @@ def monitor(suffix: str, out_file: str, exit_file: str) -> None:
             tme = time.time()
             for p in processes:
                 try:
-                    name = p.name() + "-" + suffix + "-" + str(p.pid)
+                    name = p.name()
+                    for arg in p.cmdline():
+                        if arg.endswith('.pid') or arg.endswith('.py'):
+                            name = arg.split('/')[-1].split('.')[0]
+                            break
```
putnopvut, line 43: Is there a legitimate case where there could be multiple `/` or `.` characters in the file name? If so, this could yield odd results. I tend to use Pathlib methods for getting what I want from a pathname. In this case, you can do
```
    # Assuming you have "from pathlib import PurePosixPath" in this file
    arg_path = PurePosixPath(arg)
    if arg_path.suffix == '.pid' or arg_path.suffix == '.py':
        name = arg_path.stem
        break
```
It's completely up to you if you think it's worth doing this way instead of the `split` way.

putnopvut: I've just now noticed the `[-1]` in your method, so you are accounting for multiple `/` characters. Sorry about that.

utils/process-stats.py
igsilya: I find it very difficult to have more than one positional argument with argparse.  I tried and gave up adding a required optional argument instead.

ovn-fake-multinode-utils/process-monitor.py
igsilya: No problem.  :)
In fact, we almost always have more than 1 `/`.  Current version seems to work fine with that.

utils/process-stats.py
putnopvut: Sorry for the bad suggestion, it should have been 
```
    parser.add_argument(
        'output', help='file to write an HTML result'
    )
```

igsilya: This one works.  Though I'm still not a huge fan of multiple positional arguments.
@dceara do you have a preference?

dceara: I prefer the explicit `-o` but I'm OK either way.

dceara: We discussed a bit offline, we'll go for `-o`.

PR #200, opened 2024-08-05
Fix default value for northd probe interval

igsilya:
Southbound DB cluster tolerates 16 seconds of leader being unavailable.  It doesn't make sense for northd to be more sensitive.  And also, at 500+ node scale it can take more than 5 seconds to send a full database snapshot leading to constant re-connections and 100% CPU usage on Sb DB server preparing the same snapshot over and over.

CI updated to Ubuntu 24.04, because 23.10 is EoL and images for it are no longer available on GCP.  However, CI will still fail until https://github.com/ovn-org/ovn-fake-multinode/pull/90 is merged.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ low_scale_task:
       - image_project: fedora-cloud
         image: family/fedora-cloud-38
       - image_project: ubuntu-os-cloud
-        image: family/ubuntu-2310-amd64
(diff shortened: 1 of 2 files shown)


dceara (approved):
LGTM, waiting for ovn-fake-multinode change before merging this.

dceara:
CI passed, merging, thanks!

PR #207, opened 2026-01-14
do.sh: Pin ovn-fake-multinode to v0.5.

dceara:
This is currently the latest stable tag ovn-fake-multinode provides and is tested to work with ovn-heater.  Like this we avoid unexpected breakages due to ovn-fake-multinode changes on the main branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,8 @@ cd ~/ovn-heater
 ```
 
 This step will:
-- clone OVS, OVN and ovn-fake-multinode upstream main branches in the
+- clone OVS and OVN main branches in the `runtime` directory.
+- clone ovn-fake-multinode upstream latest stable tag (`v0.5`) in the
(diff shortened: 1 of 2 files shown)


almusil (approved):

dceara:
Ah, I just noticed that this will now generate the following `config` file content for test runs:
```
ovn-fake-multinode (https://github.com/ovn-org/ovn-fake-multinode.git): d73bef8448d3c51641b018300ffecb2b806ecb4c (HEAD)
```
I'll try to change it so it generates
```
ovn-fake-multinode (https://github.com/ovn-org/ovn-fake-multinode.git): d73bef8448d3c51641b018300ffecb2b806ecb4c (v0.5)
```
if the commit is a tag instead.

almusil (approved):

PR #208, opened 2026-07-28
Migrate CI from Cirrus CI to GitHub Actions.

dceara:
Migrate ovn-heater CI from Cirrus CI (shut down June 2026) to GitHub Actions. Tests run inside QEMU/KVM VMs on ubuntu-24.04 runners with Fedora/Ubuntu matrix. VM images and runtime are cached between runs. All existing test coverage preserved. Also fixes a broken ovsdbapp install and bumps deprecated GitHub Actions to Node.js 24.

Inspired from [fafde2a8c20d](https://github.com/openvswitch/ovs/commit/fafde2a8c20d) ("github: Migrate FreeBSD CI from Cirrus CI to GitHub Actions.")
Reported-at: https://issues.redhat.com/browse/FDP-3766
Assisted-by: Claude Opus 4.6, OpenCode


.ci/ci.sh
```diff
@@ -0,0 +1,419 @@
...
+#
+# Requires OS_TYPE (fedora or ubuntu) to be set.
+
+set -o errexit
+set -x
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+TOPDIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+OS_TYPE="${OS_TYPE:?Must set OS_TYPE (fedora or ubuntu)}"
```
almusil: `OS_TYPE must be set`?

.ci/ci.sh
```diff
@@ -0,0 +1,419 @@
...
+
+# ====================================================================
+# VM helpers
+# ====================================================================
+
+VM_SSH_PORT=2222
+VM_PIDFILE=/tmp/vm.pid
+VM_NET=10.0.2.0/24
+VM_GUEST_IP=10.0.2.42
+VM_IMAGE_SIZE="40G"
```
almusil: Do we really need that big VM? Wouldn't 10G be enough?

.ci/ci.sh
```diff
@@ -0,0 +1,419 @@
...
+
+_VM_SSH_OPTS=(
```
almusil, line 29: We should add -n

.ci/ci.sh
```diff
@@ -0,0 +1,419 @@
...
+        *)
+            echo "ERROR: Unknown OS_TYPE: ${OS_TYPE}" >&2
+            exit 1
+            ;;
+    esac
+
+    local out_img="${OS_TYPE}-ci.qcow2"
+
+    echo "Downloading ${OS_TYPE} cloud image ..."
+    wget -q "${IMAGE_URL}" -O "${out_img}.dl"
```
almusil: curl? :)

.ci/ci.sh
dceara: I think I'll change it to `OS_TYPE required` to be in line with the others.

.ci/ci.sh
dceara: I can try it out.

.ci/ci.sh
dceara: While at it, I think I'll remove the "quiet" behavior in case we start failing to download images for whatever reasons.

.ci/ci.sh
dceara: Ooops, I think -n breaks rsync, e.g.:
```
+ rsync -az --delete -e 'ssh -n -p 2222 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=5 -o BatchMode=yes -o ServerAliveInterval=15 -o ServerAliveCountMax=4 -o LogLevel=ERROR -i /tmp/tmp.GpbvdU4Qj4/id_ed25519' /home/runner/work/ovn-heater/ovn-heater/ root@localhost:/root/ovn-heater/
rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: error in rsync protocol data stream (code 12) at io.c(232) [Receiver=3.2.7]
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: error in rsync protocol data stream (code 12) at io.c(232) [sender=3.2.7]
```
doing another run without -n now.

.ci/ci.sh
dceara: Ubuntu lanes sometime fail with
```
tee: /root/ovn-heater/runtime/installer-log: No space left on device
```
I'll try 20G.

.ci/ci.sh
dceara: Done.

.ci/ci.sh
dceara: Indeed, -n is not an option, I decided to not add it in the end.

.ci/ci.sh
dceara: 20G seems to be a good compromise, CI is now green.

PR #211, opened 2026-08-27
do.sh: Fix how clone_component() detects new branches/tags.

dceara:
The function didn't use to fetch tags from the origin repo and also the way it diff-ed the local code to the remote one was incorrect.  That's because:

  git diff <..> --stat | wc -l

always sets the return code to 0 (that of wc).

Instead of that, just use the more robust way of doing things

  git diff --quiet <..>


moloings1 (approved):
LGTM

dceara:
Thanks @moloings1!

